PR #2, opened 2016-05-14
dissassemble some incbins and add charmap macros for script functions

ProjectRevoTPP:
dissassemble some text incbins probably related to battle
added charmap macros for script functions and dissassemble another incbin
revise charmap.txt again
cleaned up charmap.txt and corrected incorrect info about control characters


YamaArashi:
I updated preproc so that you can write "{COLOR DARK_GREY}" instead of "{COLOR}{DARK_GREY}". Could you update this to use that format?

PR #22, opened 2016-07-06
add `dr` macro for incbinning baserom

PikalaxALT:


PikalaxALT:
apparently it conflicts with master for no reason, not going to try and fix

PR #26, opened 2016-07-29
Prof. Birch speech

camthesaxman:
This isn't ready to merge yet, but I'm just putting this here so others can review it and give suggestions. The functions marked "Fix Me" don't match the original asm code when compiled. I'm adding struct member offsets in the headers so it's easier to determine the members when decompiling.

### DIFF
--- a/src/birch_speech.c
+++ b/src/birch_speech.c
@@ -1,21 +1,32 @@
 #include "global.h"
 #include "palette.h"
 #include "sprite.h"
+#include "string_util.h"
 #include "task.h"
 #include "text.h"
 
-extern u32 gUnknown_081E768C;
-extern u32 gUnknown_081E7834;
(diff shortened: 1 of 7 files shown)


src/birch_speech.c
```diff
@@ -0,0 +1,676 @@
...
+        gTasks[taskId].func = sub_800A9EC;
+    }
+}
+
+//Fix Me
+void sub_800A9EC(u8 taskId)
+{
+    s8 n = NameMenuProcessInput();
+    
+    //The ordering of these conditional blocks in the original ROM is very strange.
```
YamaArashi: when it's like that, it means a "switch" was used

YamaArashi:
incorporated in 2085d7169e75b1fb417948d6e5167007064df258

PR #44, opened 2016-09-16
purge text label redundancy in data2

ProjectRevoTPP:


src/pokemon_2.c
```diff
@@ -21,7 +21,7 @@ extern u16 word_202FF5E;
 extern struct PokemonStorage gPokemonStorage;
 
 extern u8 gBadEggNickname[];
-extern u8 gEggNickname[];
+extern u8 gOtherText_EggNickname[];
```
YamaArashi: can you remove the "OtherText" prefix?

PR #50, opened 2016-10-02
Add DmaClear macro

camthesaxman:


YamaArashi:
This seems like a good idea. I wonder if there are other examples of DMA operations that work like this besides the wall clock. The way this macro works forces the DMA "control" value to calculated at runtime instead of compile time.

camthesaxman:
I tried replacing the DmaFill macros in main_menu.c with DmaClear, but it didn't compile the same. In wallclock.c, DmaClear works, but not DmaFill.

YamaArashi:
Yeah. This has different behavior, as I mentioned in my edits to the comment.

PR #53, opened 2016-10-03
clean up and document main_menu.c

camthesaxman:


YamaArashi:
Very nice. It's much more readable now.

PR #55, opened 2016-10-06
decompile starter chooser

camthesaxman:
This is the screen with the three Pokeballs in which the player chooses their starter Pokemon. I'm having trouble getting sub_810A42C to match, but all the rest of the code is fine. Despite having curly braces around those statements, CSE still takes effect, and the compiler just increments the pointer each time instead of recalculating the address.

### DIFF
--- a/data/data2.s
+++ b/data/data2.s
@@ -23919,15 +23919,22 @@ gBirchBallarrow_Gfx::
 gBirchCircle_Gfx::
 	.incbin "graphics/misc/birch_circle.4bpp.lz"
 
-gUnknown_083F76B8:: @ 83F76B8
-	.incbin "baserom.gba", 0x003f76b8, 0x6
+gStarterChoose_PokeballCoords:: @ 83F76B8
+	.byte 60, 64
+	.byte 120, 88
+	.byte 692, 64
 
 gUnknown_083F76BE:: @ 83F76BE
-	.incbin "baserom.gba", 0x003f76be, 0x6
+	.byte 0x00, 0x09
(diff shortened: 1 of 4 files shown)


data/data2.s
```diff
@@ -23919,15 +23919,22 @@ gBirchBallarrow_Gfx::
...
-
+	.byte 0x00, 0x09
+	.byte 0x10, 0x0A
+	.byte 0x08, 0x04
+
+gStarterMons:: @ 83F76C4
+	.2byte 277
+	.2byte 280
+	.2byte 283
+	.2byte 0
```
yenatch: Why not use `SPECIES_TREECKO` `SPECIES_TORCHIC` `SPECIES_MUDKIP`?

YamaArashi:
Doing this prevents the optimization, and puts the load of the "5" value in the right place, but the registers still don't match unfortunately.
```
    {
        u8 val = 0x5;
        {
            u8 *p = arr;
            p += 5;
            *p = 0xFC;
        }
        {
            u8 *p = arr;
            p += 6;
            *p = 0x11;
        }
        {
            u8 *p = arr;
            p += 7;
            *p = val;
        }
    }
```

data/data2.s
camthesaxman: I didn't know we had defined constants for those. Thanks, I'll add that.

camthesaxman:
The code now compiles and matches the original rom. I just have to clean it up now.

camthesaxman:
Should be ready to merge, now.

PR #58, opened 2016-10-15
decompile option menu and begin decompiling pokedex

camthesaxman:


camthesaxman:
I'm not sure how to get Task_OptionMenuSave to match. It's a small function whose main job is to store the player's options into gSaveBlock2, but the compiler is behaving strange here.

PR #62, opened 2016-10-19
write out SE as SoundEffect and BGM as BkgndMusic

camthesaxman:
It's not immediately obvious on first glance that SE stands for Sound Effect and BGM stands for Background Music, so I think this makes it a bit more readable. I usually avoid abbreviations unless the function name is over 40 characters long.


YamaArashi:
These are fairly standard abbreviations in video games and the names used in the game's sound test, so I don't see the need to change them.

camthesaxman:
I only changed the function names, like PlaySE() -> PlaySoundEffect(), which I find more descriptive, and the gMPlay_* structs. The song names are still SE_STOP, SE_KAIFUKU, BGM_TETSUJI, etc. If you think it's okay as-is, then feel free to close this.

YamaArashi:
Well, I was inspired by Pokemon X/Y symbols, such as these.

```
SEIsPlaying
BGMIsPlaying
```

I thought it would be nice to use the same terminology when it makes sense.

PR #64, opened 2016-10-20
decompile more pokedex functions

camthesaxman:
There are a few functions that I'm having trouble getting to match.


camthesaxman:
Git is being stupid right now. I'll have to create a different branch and apply my changes to that.

PR #65, opened 2016-10-20
decompile more pokedex functions

camthesaxman:
I'm having trouble getting some functions to match.


YamaArashi:
I took a look at those and yeah, they seem pretty strange. We might not be able to get everything to match.

camthesaxman:
I tried forcibly spilling the variables using &, and tried an array, and a struct, but the array seems to match the closest.

YamaArashi:
Yes. I saw those issues. I am not sure why you think the loop counter variable is a u16 in case 1 of the switch in sub_808D690. The condition branch `ble _0808D896` uses a signed comparison. It still won't match if you use s16, though.

camthesaxman:
I fixed up some mistakes to where it matches more closely. Still not 100%, though.

PR #82, opened 2016-11-02
Fixed hardcoded memory address in m4a_1.s

Diegoisawesome:


sym_bss.txt
```diff
@@ -184,7 +184,10 @@ gUnknown_030007B0: @ 30007B0
 	.space 0x4
 
 gUnknown_030007B4: @ 30007B4
-	.space 0x774
+  .space 0x4
```
YamaArashi: tab

PR #106, opened 2016-11-21
Mark header files as the correct language for GitHub

camthesaxman:
This should make Github see our header files in this repository as C instead of C++, and our .inc files as Assembly instead of Pascal and PHP.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -15,3 +15,4 @@ Makefile text eol=lf
 *.bin binary
 
 *.h linguist-language=C
+*.inc linguist-language=Assembly
--- a/.gitattributes
+++ b/.gitattributes
@@ -13,3 +13,5 @@ Makefile text eol=lf
 
 *.png binary
 *.bin binary
+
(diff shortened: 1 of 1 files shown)


NieDzejkob:
Why isn't this merged yet?

YamaArashi:
Because there is C++ code that also uses .h. Maybe we can switch to using .hpp for C++ headers.

NieDzejkob:
I think you are talking about the tools (the gamecode is pure C, right?). Maybe we could make them a separate repository, like pokered does?

yenatch:
biggest mistake pokered made

yenatch:
```
tools/agbcc/*.h linguist-language=C++
```

NieDzejkob:
AGBCC isn't even in this repository, but it's a great idea in general

YamaArashi:
I think splitting the tools and system libraries into a separate repo would be good, since Emerald uses the same tools/libraries. So I will merge this as is.

PR #111, opened 2016-11-27
Declare more non-static functions in header files

marijnvdwerf:


include/libc.h
```diff
@@ -0,0 +1,8 @@
+#ifndef GUARD_LIBC_H
+#define GUARD_LIBC_H
+
+void *memcpy(void *dest, const void *src, size_t n);
+void *memset(void *s, int c, size_t n);
+int strcmp(const char *s1, const char *s2);
+
+#endif // GUARD_LIBC_H
```
YamaArashi: Please remove this header. This is covered by the standard C headers.

marijnvdwerf:
```bash
make clean; make ruby; make compare;
...
pokeruby.gba: OK
```

camthesaxman:
I think it would be a good idea to declare functions that take no parameters with void in the parentheses. This way, the compiler is able to check that no parameters were passed.

YamaArashi:
Yeah. With the exception of GetMonData and GetBoxMonData, use `u8 sub_803DAA0(void);` instead of `u8 `sub_803DAA0();`, for example.

camthesaxman:
Const correctness would be nice, too. If a function takes a pointer and doesn't modify the object pointed to, then it should be declared const.

marijnvdwerf:
I need to include either `global.h` or `gba/types.h` in the headers for them to stop giving errors.

camthesaxman:
If you include global.h at the very beginning of all the .c files, then all of the headers included after global.h will have those type definitions.

marijnvdwerf:
@camthesaxman @YamaArashi Better?

PR #114, opened 2016-11-30
[WIP] Extract all external function declarations to headers

marijnvdwerf:
Based on #111


src/pokemon_2.c
```diff
@@ -50,12 +50,6 @@ extern u8 gUnknown_0820823C[];
 extern u8 gStatStageRatios[][2];
 extern u8 gHoldEffectToType[][2];
 
-extern u8 battle_side_get_owner(u8);
-extern u8 battle_get_side_with_given_state(u8);
-extern u32 battle_get_per_side_status(u8);
-extern u8 sub_8018324(u8, u8, u8, u8, u16);
-extern u8 sub_803C348(u8);
-
 u8 sub_803C348(u8 a1)
```
marijnvdwerf: Has two errors in the generated pokemon_2.s:

(left is `master`, right is this PR)
image
image

marijnvdwerf:
This gets rid of all `implicit declaration of function '{{ name}}'` warnings.

marijnvdwerf:
Two issues remaining in `sub_803C434 `:
image
image

YamaArashi:
Yeah. That function needed a lot of hacks to get it to match. I suspect there wouldn't be a problem if the function were written correctly. I'm not sure if there's any solution but to change the function back to asm.

PR #118, opened 2016-12-03
Fix various warnings

marijnvdwerf:
Remaining warnings:
```txt
src/battle_setup.c: In function `GetTrainerFlagFromScriptPointer':
src/battle_setup.c:690: warning: cast to pointer from integer of different size
src/weather.c: In function `sub_807CB10':
src/weather.c:170: warning: assignment from incompatible pointer type
src/weather.c:172: warning: assignment from incompatible pointer type
```

### DIFF
--- a/include/asm.inc.h
+++ b/include/asm.inc.h
@@ -266,6 +266,11 @@ void sub_80A68CC();
 
 // asm/contest.o
 void sub_80AB1B0(void);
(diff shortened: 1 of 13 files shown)


ProjectRevoTPP:
Refrain from this file until I finish decompiling it. I don't want a ton of conflicts

PR #147, opened 2017-01-02
No more baserom in data/graphics.s

yenatch:


marijnvdwerf:
I've identified/named some of these already for the German rom: https://github.com/marijnvdwerf/pokeruby/tree/german-2

PR #171, opened 2017-01-07
Continue decompilation of credits

marijnvdwerf:
https://www.youtube.com/watch?v=jNViqLuacWo


camthesaxman:
Can you make the code style and formatting consistent with the rest of the repository? We use Allman/BSD-style bracing, and do not typedef structs. Also, switch statements should have the case labels aligned with the brace like this:
```
switch (foo)
{
case 0:
     DoSomething();
     break;
case 1:
    DoSomethingElse();
    break;
}
```

PR #177, opened 2017-01-08
start decompiling field_control_avatar.c

camthesaxman:


src/field_control_player_avatar.c
```diff
@@ -0,0 +1,2 @@
+#include "global.h"
```
YamaArashi: what is this?

camthesaxman: I'm including global.h like almost all of our other C files do.

YamaArashi: what is this file?

camthesaxman: Oh, that file. I accidentally named it the wrong thing and forgot to delete it.

PR #198, opened 2017-01-18
start decompiling rom3

camthesaxman:


src/rom3.c
```diff
@@ -0,0 +1,435 @@
...
+        gUnknown_02024A72[i] = 0xFF;
+        gUnknown_02024E60[i] = 0;
+        gUnknown_02024E64[i] = 0;
+    }
+    sub_800B858();
+    gUnknown_02024A64 = 0;
+    battle_anim_clear_some_data();
+    sub_8040710();
+    BattleAI_HandleItemUseBeforeAISetup();
+    if (gBattleTypeFlags & 0x10)
```
YamaArashi: `if (gBattleTypeFlags & BATTLE_TYPE_FIRST_BATTLE)`

This function checks if the current battle is the Poochyena battle, but it may be used in other contexts as well.

camthesaxman: What should I name it?

YamaArashi: I'm not sure. I guess we can just rename it when we found out more about the structure of the battle system.

PR #199, opened 2017-01-19
decompile more of rom3

camthesaxman:
Why does this conflict? Nothing has changed in master since my last pull request using this branch.


sh.exe.stackdump
```diff
@@ -0,0 +1,16 @@
+Exception: STATUS_STACK_OVERFLOW at rip=00077EF25B7
```
YamaArashi: You included this by accident.

PR #200, opened 2017-01-19
[Don't merge] Introduce clang-format

marijnvdwerf:
Still figuring out the proper format. Please mark desired/undesired changes this would cause.


marijnvdwerf:
So, `clang-format` doesnt like the `_"..."` used for marking strings that need to be converted. CLion also chokes on it locally for me. Is there a way to get it to use a function-like macro that makes it valid C?

YamaArashi:
I could modify it to use `_("...")` instead of `_"..."`. This tool seems to mess up a lot of the formatting though.

marijnvdwerf:
Yeah, I'm trying to decrease the number of differences, but there are a lot of styles all over the project. I think the general idea of `clang-format` is that you have CI run it on the commit it's testing, so you get feedback when the code you added/changed doesn't match the coding style guide.

PR #224, opened 2017-01-29
set up Travis CI

camthesaxman:


camthesaxman:
Ok, this should be ready to merge. I've used a secure environment variable for the baserom.gba download URL so that Travis can obtain a copy of the baserom without sharing the link. You will need to regenerate that after this is merged since the current one is tied to my fork.

PR #226, opened 2017-01-30
set up Travis CI

camthesaxman:
Ok, this should be ready to merge. I've used a secure environment variable for the baserom.gba download URL so that Travis can obtain a copy of the baserom without sharing the link. ~~You will need to regenerate that using the travis ruby gem after this is merged since the current key is tied to my fork.~~
All you need to do to get Travis up and running is to grant Travis access to this repository, and in the project settings, add an envrionment variable called `BASEROM_URL` containing a download link to the baserom. Any link that can be downloadable with wget will work. (PM me on Slack if you need one.) After that, it should just work, and we'll have automatic builds to make sure the ROMs match each time someone submits a pull request.
travis_env_var


YamaArashi:
handled by #258

PR #236, opened 2017-02-19
Some characters went missing in the previous PR

Wack0:
...whoops >_>


YamaArashi:
Ah. I didn't see this PR. I fixed this in c0afef324418f190cfbc6d6b3447e660479b4aeb.

Wack0:
And there was another character that went missing that you noticed and I didn't.

I still don't know how they even went missing in the first place.. :/

YamaArashi:
It looks like you somehow deleted the characters in your second commit to fix the indentation.

PR #239, opened 2017-02-28
Add German graphics

marijnvdwerf:


YamaArashi:
Is this ready to merge?

marijnvdwerf:
Yup. These graphics build a matching German rom, and the palettes match. Except for `trade/menu.png`, but I believe my palette is the right one

PR #242, opened 2017-03-23
Build multiple versions at a time.

yenatch:
`make tidy` before merging locally.


ProjectRevoTPP:
NOTE: This does not work on older GNU Make versions. Please use 4.x after this update.

PR #248, opened 2017-04-27
Sort includes

marijnvdwerf:
Sorted as following:
```
"global.h"
"gba/gba.h"
"gba/*"
"includes/{{ self }}.h"
"includes/*"
```


ProjectRevoTPP:
There's so many includes per file. Could we get some sort of superheaders soon that combine a bunch of similar headers into one file? Like for example `field.h` includes all the minor field headers, `time.h` includes `rtc.h` and `clock.h`, etc.

PR #271, opened 2017-05-05
German ROM: berry.c

marijnvdwerf:


YamaArashi:
How can Travis check builds when the German Ruby won't have the correct hash for a while?

marijnvdwerf:
It should only validate whether the German rom actually builds. Having the decompiled stuff in is easier than trying to keep it in sync.

YamaArashi:
OK. I think we need to add a new phony target for German builds, in order to set the language properly.

.travis.yml
```diff
@@ -28,3 +28,4 @@ script:
   - make -j2 compare_sapphire
   - make -j2 compare_sapphire_rev1
   - make -j2 compare_sapphire_rev2
+  - make -j2 pokeruby_de.gba
```
YamaArashi: make -j2 ruby_de

PR #292, opened 2017-05-11
Add stub Cmake project for IDE integration

marijnvdwerf:
Building doesn't work, but this does fix include detection for IDEs like CLion.


camthesaxman:
I don't think it's worth it to include this in the repo, unless most contributors are using an IDE that needs this. You can keep this file in your local repository without it being tracked by Git if you add it to your local `.git/info/exclude` which has the same format as `.gitignore`.

PR #293, opened 2017-05-12
add LDFLAGS variable to tools makefiles

camthesaxman:
This is mainly to make it easier to build redistributable Windows binaries. Running `LDFLAGS=-static ./build_tools` will build static versions of the tools that don't depend on any external DLLs.


yenatch:
Are there any drawbacks to `LDFLAGS += -static ./build_tools` being the default?

camthesaxman:
It didn't build correctly on Linux when I did that. `libpng` pulls in some extra dependencies when I statically link it on Linux.

PR #295, opened 2017-05-12
Decompile/credits 3

marijnvdwerf:


src/credits.c
```diff
@@ -1639,15 +1512,12 @@ void sub_8143648(u16 paletteTag, u8 arg1);
...
-    u32 species;
     u32 personality;
     void *palette;
     u8 spriteId;
     u8 spriteId2;
 
-    // FIXME: For some reason r0 is copied to r6 before.
-    species = NationalPokedexNumToSpecies(nationalNum);
-
+    species = NationalPokedexNumToSpecies(species);
```
camthesaxman: `species` has not been initialized here when it is passed to `NationalPokedexNumToSpecies`. It matches if you completely replace `species` with nationalNum. I'm not sure what to rename the variable, though, since it is reused for both purposes.

PR #299, opened 2017-05-15
Secret base and TV decompilation

PikalaxALT:


ld_script.txt
```diff
@@ -174,6 +174,7 @@ SECTIONS {
         src/record_mixing.o(.text);
         asm/record_mixing.o(.text);
         asm/sound_check_menu.o(.text);
+		src/secret_base_80bb4ac.o(.text);
         asm/secret_base.o(.text);
         src/secret_base.o(.text);
```
marijnvdwerf: make this `src/secret_base.o(.text_b);`, and then add this line before the function already in there: `asm(".section .text_d");`. That way you can just put your stuff in secret_base.c

src/secret_base_80bb4ac.c
```diff
@@ -0,0 +1,198 @@
...
+        return 5;
+    else if (v0 == 0x98 || v0 == 0x99)
+        return 6;
+    return 0;
+}
+#else
+__attribute__((naked))
+u8 sub_80BB66C(void) {
+    asm(".syntax unified\n\
+    	push {r4,lr}\n\
```
marijnvdwerf: tabs

src/secret_base.c
```diff
@@ -2,10 +2,831 @@
...
+extern u16 gScriptResult;
+extern const struct
+{
+    u16 unk_083D1358_0;
+    u16 unk_083D1358_1;
+} gUnknown_083D1358[7];
+extern const u8 gUnknown_083D1374[4 * 16];
+extern void *gUnknown_0300485C;
+extern u8 sub_807D770(void);
+extern u8 gUnknown_083D13EC[12];
```
camthesaxman: Please declare this as `const` along with any other variables that are stored in ROM (ROM variables have an 08XXXXXX address).

PikalaxALT: I cannot declare gUnknown_081A2E14 as a const: `passing arg 1 of `ScriptContext1_SetupScript' discards qualifiers from pointer target type`

PR #306, opened 2017-05-22
De/integrate 3

marijnvdwerf:


yenatch:
The review didn't send any notifications so I'm commenting in case you didn't see it.

marijnvdwerf:
Yeah I get what you're saying. When I started, I was planning to to something like `#define GLYPH_AB GLPYH_G_CAP`. Most width changes don't seem to make that much sense though (I expected them to line up with the font changes).

Here are the different glyphs: https://github.com/marijnvdwerf/pokeruby/pull/7

I'm just gonna remove the commit for now.

PR #309, opened 2017-05-25
item.c: Replace item identifier with item enum

Wack0:
CopyItemName() hardcoded the Enigma Berry identifier, replaced it with ITEM_ENIGMA_BERRY.

It still hardcodes berry identifier 0x2B (Enigma Berry), but that's because I can't seem to find a similar enum for berries.


yenatch:
Feel free to add one. There are lots of places that use the berry id.

PR #315, opened 2017-06-03
Improvements for Mid2Agb

easyaspi314:
Some improvements for Mid2Agb:

* Better support for drag and drop
* Support for filenames with spaces. This is also important for drag and drop because otherwise the ASM label would be something like "C:\Users\easyaspi314\pokeruby\tools\mid2agb\my midi file with spaces", and you can't have spaces in labels.
* Support for gba_mus_riper looping. gba_mus_riper uses loopStart and loopEnd instead of [ and ].
* Reveal -L option in usage. For the reference, it sets the ASM label.
* Header in usage message with program name and copyright.
* Slightly better grammar in error messages. Mainly just capitalization
* Pause on error. This allows you to actually see the error if you drag and drop. You need to press Enter to close it now.
* Not mentioned in the commit, but I added a few comments.

### DIFF
--- a/tools/mid2agb/error.cpp
+++ b/tools/mid2agb/error.cpp
@@ -22,7 +22,8 @@
 #include <cstdlib>
 #include <cstdarg>
 
-// Reports an error diagnostic and terminates the program.
+// Reports an error diagnostic and terminates the program
+// after the user presses Enter.
 [[noreturn]] void RaiseError(const char* format, ...)
 {
     const int bufferSize = 1024;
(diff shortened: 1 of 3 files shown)


yenatch:
Pause on error is undesirable in a build. Can drag and drop be detected?

easyaspi314:
I'd have to look into it.

easyaspi314:
In the build system, could we maybe put a flag on by default? Like -f to build despite errors?

marijnvdwerf:
Not familiar with Windows CLI practices, but wouldn't it make more sense to have something like a .bat file as a droplet?

easyaspi314:
That might work.

easyaspi314:
Also, unrelated, but is there a specific reason these aren't in pokeemerald?

Why don't we put this in pokeruby-tools and have it be the precompiled .exes in their own folder; then link pokeruby and pokeemerald to it?

These are really useful for all GBA games, not just Ruby.

camthesaxman:
Ideally, our mid2agb should have the exact same behavior as the one from the official SDK, and should be a drop-in replacement.

PikalaxALT:
This PR may be obsolete, please review.

PR #325, opened 2017-06-13
Clean/clang format upstream

marijnvdwerf:


include/items.h
```diff
@@ -1,388 +1,389 @@
 #ifndef GUARD_ITEMS_H
 #define GUARD_ITEMS_H
 
-enum {
```
yenatch: weird (bad)

include/rtc.h
```diff
@@ -3,19 +3,19 @@
 
 #include "siirtc.h"
 
-#define RTC_INIT_ERROR         0x0001
```
marijnvdwerf: Add guard

include/pokemon.h
```diff
@@ -1,125 +1,126 @@
 #ifndef GUARD_POKEMON_H
 #define GUARD_POKEMON_H
 
-#define MON_DATA_PERSONALITY        0
```
marijnvdwerf: add guard

include/items.h
marijnvdwerf: What specific part are you talking about?

yenatch: It aligns comments to each block of comments but it should align to the whole block of code.

PR #328, opened 2017-06-16
Pokémon data

PikalaxALT:
Decompile and delete data/pokemon.s and its unique dependents


src/pokemon_data.c
```diff
@@ -931,3 +931,422 @@ const s8 gNatureStatTable[][5] = {
...
+    {0xc79b4621, 0x00a00ed2}, // 0b0010100000000011101101001011000111100110110100011000100001 (Registeel)
+    {0xc79b727c, 0x03b00e42}, // 0b1110110000000011100100001011000111100110110111001001111100 (Kyogre)
+    {0xcff946b2, 0x00a60ef6}, // 0b0010100110000011101111011011001111111110010100011010110010 (Groudon)
+    {0xc7f376b6, 0x03ba0eb6}, // 0b1110111010000011101011011011000111111100110111011010110110 (Rayquaza)
+    {0xb7bbd63e, 0x035c5e93}, // 0b1101011100010111101001001110110111101110111101011000111110 (Latias)
+    {0xb7bbd63e, 0x035c5e93}, // 0b1101011100010111101001001110110111101110111101011000111110 (Latios)
+    {0xb59bc62c, 0x00408e93}, // 0b0001000000100011101001001110110101100110111100011000101100 (Jirachi)
+    {0xf5bbde2d, 0x00e58fc3}, // 0b0011100101100011111100001111110101101110111101111000101101 (Deoxys)
+    {0xb41b8e28, 0x00419f03}, // 0b0001000001100111110000001110110100000110111000111000101000 (Chimecho)
+};
```
yenatch: binary literals may be a good preproc feature

src/pokemon_data.c
```diff
@@ -1350,3 +1351,301 @@ const u32 gTMHMLearnsets[][2] = {
...
+    TRAINER_CLASS_NAME_MAGMA_LEADER,      // MAGMA_LEADER
+    TRAINER_CLASS_NAME_TEAM_MAGMA,        // TEAM_MAGMA_M
+    TRAINER_CLASS_NAME_TEAM_MAGMA,        // TEAM_MAGMA_F
+    TRAINER_CLASS_NAME_LASS,              // LASS
+    TRAINER_CLASS_NAME_BUG_CATCHER,       // BUG_CATCHER
+    TRAINER_CLASS_NAME_HIKER,             // HIKER
+    TRAINER_CLASS_NAME_YOUNG_COUPLE,      // YOUNG_COUPLE
+    TRAINER_CLASS_NAME_OLD_COUPLE,        // OLD_COUPLE
+    TRAINER_CLASS_NAME_SIS_AND_BRO        // SIS_AND_BRO
+};
```
yenatch: probably multiple files

src/pokemon_data.c
marijnvdwerf: Why not `TM01 | TM02 | TM03`? Also, is there a reason we can't use `u64`s for this?

data/spinda_spot_graphics.inc
```diff
@@ -1,81 +0,0 @@
-	.align 2
-gSpindaSpotGraphics:: @ 81FCB84
-	.byte 16 @ X
-	.byte 14 @ Y
-	.2byte 0b0000000001110000
```
marijnvdwerf: Doesn't gbagfx also support 1bpp graphics?

src/pokemon_data.c
yenatch: If you changed the move, you would have to set the TM01 flag for all Pokemon, which is error prone.

```c
struct TMLearn {
    u64 FocusPunch:1;
}
const struct TMLearn gTMLearns = {
    /* Machamp */ {
        .FocusPunch=1,
    }
};
```

If you did this:
```diff
 struct TMLearn {
-    u64 FocusPunch:1;
+    u64 HyperBeam:1;
 }
```

You would get compile errors on Pokemon that learn Focus Punch, and you wouldn't have to worry about Pokemon that don't. You also don't run the risk of any Pokemon learning Hyper Beam when they shouldn't. Also, you don't have to explicitly say that a Pokemon does not learn a move (but you can if you like). This behavior is very desirable.

yenatch: Agreed on u64.

include/pokemon.h
```diff
@@ -407,6 +407,11 @@ struct SpindaSpot
     u16 image[16];
 };
 
+struct __attribute__((packed)) LevelUpMove {
+    u16 move:9;
+    u16 level:7;
+};
+
```
yenatch: dead code. i take it this doesn't match? i think you have to specify packed in the array declaration if that's giving you trouble.

marijnvdwerf:
Wouldn't dot-notation make more sense for `BaseStats`?

yenatch:
There isn't really much of a difference. I did up an example and it wasn't any clearer than before.

src/pokemon_data.c
```diff
@@ -15907,3 +15905,603 @@ const u16 gChimechoLevelUpLearnset[] = {
...
+    }}, {{ // Registeel
+    }}, {{ // Kyogre
+    }}, {{ // Groudon
+    }}, {{ // Rayquaza
+    }}, {{ // Latias
+    }}, {{ // Latios
+    }}, {{ // Jirachi
+    }}, {{ // Deoxys
+    }}, {{ // Chimecho
+    }}
```
yenatch: What are the double braces for? At first I thought it was some clever way to avoid `{0}` termination but it's fixed length.

PikalaxALT: Artifact of semiautomation

src/calculate_base_damage.c
```diff
@@ -41,7 +41,7 @@ extern struct BaseStats gBaseStats[];
 extern struct SpriteTemplate gSpriteTemplate_8208288[];
 extern u8 gTrainerClassToPicIndex[];
 extern u8 gTrainerClassToNameIndex[];
-extern u8 gSecretBaseTrainerClasses[];
+// extern u8 gSecretBaseTrainerClasses[];
```
yenatch: ?

src/pokemon_data.c
PikalaxALT: The question is whether using `u64` would break the function that retrieves TM/HM learns, which expects a `u32[2]`.

src/calculate_base_damage.c
PikalaxALT: I think I did that because it's either not used, or it's defined in an include.

PikalaxALT: i commented to make it easy to undo if needed.  Turns out, undo was not needed.

include/pokemon.h
PikalaxALT: Yeah, it currently doesn't match at point of use.  I haven't been able to fix that.  In the event that someone else can, I leave them this struct.

data/spinda_spot_graphics.inc
PikalaxALT: Maybe.  Are you suggesting I do this as INCBIN_U16 calls?

src/pokemon_data.c
PikalaxALT: yeah, probably should use includes similar to field_map_obj_data

src/pokemon_data.c
PikalaxALT: To answer that question: yes it does.

include/data/pokemon/tmhm_learnsets.h
```diff
@@ -0,0 +1,427 @@
...
+    {0xcff946b2, 0x00a60ef6}, // 0b0010100110000011101111011011001111111110010100011010110010 (Groudon)
+    {0xc7f376b6, 0x03ba0eb6}, // 0b1110111010000011101011011011000111111100110111011010110110 (Rayquaza)
+    {0xb7bbd63e, 0x035c5e93}, // 0b1101011100010111101001001110110111101110111101011000111110 (Latias)
+    {0xb7bbd63e, 0x035c5e93}, // 0b1101011100010111101001001110110111101110111101011000111110 (Latios)
+    {0xb59bc62c, 0x00408e93}, // 0b0001000000100011101001001110110101100110111100011000101100 (Jirachi)
+    {0xf5bbde2d, 0x00e58fc3}, // 0b0011100101100011111100001111110101101110111101111000101101 (Deoxys)
+    {0xb41b8e28, 0x00419f03}, // 0b0001000001100111110000001110110100000110111000111000101000 (Chimecho)
+};
+
+#endif //POKERUBY_TMHM_LEARNSETS_H
```
yenatch: This is harder to modify than before. Until the rewrite we should stick with `0b`.
```c
asm(
    ...
    ".8byte 0b0001000001100111110000001110110100000110111000111000101000" // Chimecho
);
```

PR #334, opened 2017-06-19
reduce verbosity of build (add VERBOSE option to makefile)

camthesaxman:


yenatch:
`$Q` is somewhat idiomatic

camthesaxman:
I was hoping to reduce the log size for Travis, but it seems that it does it by line count. I guess this isn't really that useful.

PR #358, opened 2017-07-08
Build German Sapphire (1.0)

marijnvdwerf:


asm/crt0.s
```diff
@@ -47,10 +47,19 @@ GPIOPortReadEnable: @ 80000C8
 	.4byte 0xFFFFFFFF
 	.4byte 0xFFFFFFFF
 	.4byte 0xFFFFFFFF
+	.ifdef SAPPHIRE
+	.4byte          1
+	.else
 	.4byte          2
+	.endif
```
yenatch: this is probably `VERSION_SAPPHIRE`/`VERSION_RUBY`

yenatch: this is probably `GAME_VERSION`

asm/crt0.s
```diff
@@ -47,10 +47,19 @@ GPIOPortReadEnable: @ 80000C8
...
 	.4byte          5
```
marijnvdwerf: Likely GAME_LANGUAGE

PR #360, opened 2017-07-12
(don't merge yet) convert map data to C

camthesaxman:


camthesaxman:
This won't work. I still need to think about how I'll organize the map data.

PR #362, opened 2017-07-13
Decompile menu_helpers

drifloony:


src/menu_helpers.c
```diff
@@ -0,0 +1,478 @@
...
+    else
+    {
+        MenuPrint(gStringVar1, left, top);
+    }
+}
+
+// unused
+void PrintTriangleCursorWithPalette(u8 left, u8 top, u8 paletteNum)
+{
+    u8 buffer[2];
```
marijnvdwerf: These `memcpy`s should be fairly trivial to replace with `u8 buffer[] = _("▶$")`; That should place the string in the `.rodata` section of the object file.

yenatch: What about `PrintStringWithPalette(_("▶"), paletteNum, left, top);`?

marijnvdwerf: It might, but it probably won't work in `PrintStringWithPalette`

camthesaxman: You could do `PrintStringWithPalette(&(u8[])_("▶"));`, but that's a bit ugly, and I think the way marijnvdwerf suggested is fine. Just rename `buffer` to something useful like `rightArrow`.

yenatch: should `preproc` be doing this cast instead?

marijnvdwerf: (Check the actual function `PrintStringWithPalette` in this PR to see what I meant)

PR #364, opened 2017-07-14
finish decompiling record_mixing

camthesaxman:


src/record_mixing.c
```diff
@@ -25,33 +25,46 @@
...
+// I can't define these here or else RecordMixing_PrepareExchangePacket gets optimized.
+/*
+void *const recordMixingSecretBases = &gSaveBlock1.secretBases;
+void *const recordMixingTvShows = &gSaveBlock1.tvShows;
+void *const gUnknown_083D0274 = &gSaveBlock1.unknown_2ABC;
+void *const gUnknown_083D0278 = &gSaveBlock1.oldMan;
+void *const recordMixingEasyChatPairs = &gSaveBlock1.easyChatPairs;
+struct RecordMixing_UnknownStruct *const gUnknown_083D0280 = &gUnknown_02038738;
+void *const gUnknown_083D0284 = &gSaveBlock2.filler_A8;
+*/
```
yenatch: Is there a way around this? What if it's an array of `void *const`?

camthesaxman: No, that doesn't match. I also tried a struct, and it does the same thing. What does match is if I uncomment those variables and remove the `const`. However, that doesn't build because it gets put into `.data` instead of `.rodata`. I have a feeling that the original linker script put `.data` into ROM along with `.rodata`, and Game Freak didn't use `const` on every single piece of ROM data.

drifloony: It's possible that they had something like `*(.data .rodata)` since I doubt they would have listed each file individually in the linker script.

camthesaxman: That seems likely. As long as ld links files in the same order that they were passed, we can specify them in the Makefile and use that.

PR #370, opened 2017-07-16
map out ram variables

drifloony:
I moved RAM variables into C files when possible and put comments for files that have not yet been decompiled.


src/save_failed_screen.c
```diff
@@ -22,17 +22,11 @@
...
-    u16 timer; // appears to be unused, the only case its "used" cannot be reached normally
-    u8 unused[0xE]; // appears to be unused space. alternatively, there could have been multiple clock structs. however, neither of these cases are fulfilled, so the purpose of this space is unknown. one theory is that there is room for 3 more clock structs (2 16-bit values), so its possible GF intended there to be support for link-based save failed screens in case the synchronization failed, or it in fact was present in an earlier build but taken out for reasons.
-};
-
 extern u8 unk_2000000[];
 
-extern u16 gSaveFailedType;
-extern struct SaveFailedClockStruct gSaveFailedClockInfo;
+static EWRAM_DATA u16 gSaveFailedType = 0;
+static EWRAM_DATA u16 gSaveFailedClockInfo[9] = {0};
```
camthesaxman: Why was this changed to a u16 array?

src/pokeblock.c
```diff
@@ -26,8 +26,15 @@
 #include "songs.h"
 #include "safari_zone.h"
 #include "use_pokeblock.h"
+#include "event_data.h"
 #include "pokeblock.h"
 
+extern u8 ewram[];
+
+static EWRAM_DATA u8 gUnknown_02039244 = 0;
+static EWRAM_DATA ALIGNED(4) u8 gUnknown_02039248[4] = {0}; // why aligned to 4 bytes?
```
camthesaxman: This seems like it would be a struct with four u8 fields, if the code still matches with that.

src/save_failed_screen.c
drifloony: It's not aligned to 4 bytes. Putting the gSaveFailedType variable into the struct didn't match either.

PR #373, opened 2017-07-17
decompile pokemon_icon

drifloony:


src/pokemon_icon.c
```diff
@@ -0,0 +1,1276 @@
...
+};
+
+const struct SpritePalette gMonIconPaletteTable[] =
+{
+    { &gMonIconPalettes[0][0], POKE_ICON_BASE_PAL_TAG + 0 },
+    { &gMonIconPalettes[1][0], POKE_ICON_BASE_PAL_TAG + 1 },
+    { &gMonIconPalettes[2][0], POKE_ICON_BASE_PAL_TAG + 2 },
+
+// These palettes are actually part of the title screen BG palette and are not
+// used for Pokémon icons.
```
camthesaxman: You can also use `gMonIconPalettes[n]` here instead of `&gMonIconPalettes[n][0]`.
These out of bounds pointers are actually part of `gUnknown_08E96738`, which is used by pokedex.c.

drifloony: That comment was copied from another file. I think it refers to Emerald rather than Ruby/Sapphire. https://github.com/pret/pokeemerald/blob/4db33778ad3faa64d994f46358a28c22c393f7c6/data/graphics.s#L1017

src/pokemon_icon.c
```diff
@@ -15,16 +16,1006 @@ struct MonIconSpriteTemplate
...
+
+static const union AffineAnimCmd sAffineAnim_0[] =
+{
+    AFFINEANIMCMD_FRAME(0, 0, 0, 10),
+    AFFINEANIMCMD_END,
+};
+
+static const union AffineAnimCmd sAffineAnim_1[] =
+{
+    AFFINEANIMCMD_FRAME(0xFFFE, 0xFFFE, 0, 122),
```
camthesaxman: I think this would be better written as `AFFINEANIMCMD_FRAME(-2, -2, 0, 122),` since those fields are signed.

drifloony: Well, these are actually fixed-point numbers. The value is really -2/256 or -0.0078125 in decimal. I'm not sure what the best way to represent these is.

drifloony: I thought these values were written directly into the hardware registers, but there is some additional scaling done before that. I'm not sure what kind of effect this value really has.

camthesaxman: Yeah, looks like it ends up being 0x10000 / xScale, then is passed as a Q8.8 to the ObjAffineSet syscall. I think using -2 is fine here, and the other files use negative parameters for this. We might consider using a fixed-point macro in the future like trig.c does.

PR #382, opened 2017-07-21
isolate German changes

camthesaxman:
This PR merges the `-de` directories into the English ones so that both the English and German ROMs can be built without copying and overwriting files.

This does introduce a lot of `.ifdef`s. Not sure if that's acceptable.


DizzyEggg:
I don't like the idea tbh. The code becomes harder to read and makes searching for things difficult.

camthesaxman:
Yeah, I'll probably close this eventually. I'm trying to set this up where preproc and scaninc search different paths for the German and English version, but I'm struggling to get that to work.

PR #388, opened 2017-07-24
[dont merge] scaninc: read C includes and add -I

yenatch:
This also makes scaninc faster than before despite the extra overhead of c includes.


yenatch:
This can be merged when a build is added to pokeruby-tools.

tools/scaninc/c_file.cpp
```diff
@@ -272,16 +290,7 @@ void CFile::CheckIncbin()
...
-
-    SkipWhitespace();
-
-    if (m_buffer[m_pos] != ')')
-        FATAL_INPUT_ERROR("expected ')'");
-
     m_pos++;
 
-    m_incbins.emplace(path);
+    return std::string(m_buffer, startPos, m_pos - 1 - startPos);
```
camthesaxman: This is a buffer overflow because `m_buffer` is not null terminated when it gets implicitly converted to a `std::string` object. Use `std::string(m_buffer + startPos, m_pos - 1 - startPos);` instead.

camthesaxman:
Any plans on when you want this to be merged?

yenatch:
It needs to be in pokeruby-tools first since it behaves differently without -I than before.

PR #393, opened 2017-07-31
decompile battle_message.c

DizzyEggg:
There is one non-matching function. It's a switch that appends various strings to the destination passed as argument. It's very small function. What's interesting is that ALL of the strings are just exclamation marks. I guess the strings are different in various language versions?

I didn't add any rodata, only decompiled functions. I tried to but I decided to leave it someone more experienced due to the English/German differences.

I also didn't rename any of the global variables, because there were simply too many of them. I hope someone will clean this all up in the future.


camthesaxman:
Can you fix the German build? If you want an `ifdef`d version of the asm so you can see what the German changes are, look here. https://github.com/camthesaxman/pokeruby/blob/integrate_german/asm/battle_message.s

DizzyEggg:
I don't know, don't really feel like fixing the German build. Let's wait for marijn to clean it up later.

camthesaxman:
I don't know if he's coming back soon. If you don't feel like decompiling it, you can just put the German code as inline assembly under an ifdef, and I'll fix it later. We at least need it to build correctly before I merge this.

PR #398, opened 2017-08-01
decompile hall of fame

DizzyEggg:
Function, variable and struct name suggestions are welcome.


src/hall_of_fame.c
```diff
@@ -99,7 +146,7 @@ bool8 sub_8141E64(void)
         UpdatePaletteFade();
         if (!gPaletteFade.active)
         {
-            SetMainCallback2(sub_8141E4C);
+            SetMainCallback2(CB2_HallOfFame);
             PlayBGM(436);
```
camthesaxman: `PlayBGM(BGM_DENDOU);`

DizzyEggg: Fixed plus added a file static variable

PR #399, opened 2017-08-02
Improve battle_4

Doesnty:
Only function I was able to get matching was TypeCalc, but hopefully the adjustments to other non-matching functions to make them buildable and stab at atk46_playanimation2 will make things easier on someone else.

### DIFF
--- a/src/battle_4.c
+++ b/src/battle_4.c
@@ -93,6 +93,7 @@ extern u16 gUnknown_02024C2C[4]; //last used moves 2, used by sketch
 extern u16 gUnknown_030041B0;
 extern u16 gUnknown_02024C4C[4]; //last used moves by banks, another one
 extern u8 gCurrentMoveTurn;
+extern u16 gTrappingMoves[];
 
 //extern functions
 u8 AtkCanceller_UnableToUseMove(void);
@@ -199,6 +200,9 @@ extern u8 BattleScript_TrainerBallBlock[];
 extern u8 BattleScript_WallyBallThrow[];
 extern u8 BattleScript_SuccessBallThrow[];
 extern u8 BattleScript_ShakeBallThrow[];
+extern u8 BattleScript_AllStatsUp[];
+extern u8 BattleScript_AtkDefDown[];
+extern u8 BattleScript_SAtkDown2[];
 
 extern u8 gUnknown_081D919F[]; //spikes1
 extern u8 gUnknown_081D9171[]; //spikes2
@@ -417,6 +421,7 @@ static void atk85_stockpile(void);
 static void atk86_stockpiletobasedamage(void);
 static void atk87_stockpiletohpheal(void);
 static void atk88_negativedamage(void);
+static u8 ChangeStatBuffs(s8, u8, u8, u8*);
 static void atk89_statbuffchange(void);
 static void atk8A_normalisebuffs(void);
 static void atk8B_setbide(void);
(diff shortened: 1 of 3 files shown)


src/battle_4.c
```diff
@@ -6003,12 +5767,12 @@ static void atk23_getexp(void)
...
-                BATTLE_STRUCT->beforeLvlUp[2] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_DEF);
-                BATTLE_STRUCT->beforeLvlUp[3] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_SPD);
-                BATTLE_STRUCT->beforeLvlUp[4] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_SPATK);
-                BATTLE_STRUCT->beforeLvlUp[5] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_SPDEF);
+                ((u16*)ewram + 0xB8C0)[0] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_MAX_HP); //doesnt match
+                ((u16*)ewram + 0xB8C0)[1] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_ATK);
+                ((u16*)ewram + 0xB8C0)[2] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_DEF);
+                ((u16*)ewram + 0xB8C0)[3] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_SPD);
+                ((u16*)ewram + 0xB8C0)[4] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_SPATK);
+                ((u16*)ewram + 0xB8C0)[5] = GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_SPDEF);
```
yenatch: what is improved about this

Doesnty: it comes closer to matching, iirc

i do agree that it looks stupid, should probably have put the old lines in a comment

src/battle_4.c
```diff
@@ -5927,7 +5375,7 @@ static void atk23_getexp(void)
 
                 if (GetMonData(&gPlayerParty[BATTLE_STRUCT->expGetterID], MON_DATA_HP))
                 {
-                    s16 stringID;
+					s16 stringID;
```
camthesaxman: tab

src/battle_4.c
```diff
@@ -1671,7 +1680,7 @@ u8 TypeCalc(u16 move, u8 bank_atk, u8 bank_def)
                     ModulateDmgByType2(gTypeEffectiveness[i + 2], move, &flags);
                 //check type2
                 if (gTypeEffectiveness[i + 1] == gBattleMons[bank_def].type2 &&
-                    gBattleMons[bank_def].type1 != gBattleMons[bank_def].type2)
+                    gBattleMons[gBankTarget /* what the christ */].type1 != gBattleMons[bank_def].type2)
```
camthesaxman: What's so unusual about this?

Doesnty: It's supposed to be checking to see if the mon is single-type or not by seeing if its types are the same, but isn't indexing with the same variable. It happens to work fine to my knowledge because bank_def should always be the same as gBankTarget but I would not be surprised if there's some obscure AI bug because of this.

PR #403, opened 2017-08-12
split out src/ directory into categorized subdirectories.

ProjectRevoTPP:
Please leave comments as to what files should go where, but I think this is a decent starting point.


camthesaxman:
libc, m4a, siirtc, and agb_flash don't really belong in engine, since those are SDK provided libraries, and not part of Pokemon Ruby. Having a libs directory for those would probably be a good idea.

ProjectRevoTPP:
I tried to put them in lib/ and couldnt figure out how to do it.

camthesaxman:
How is that any different from putting it in the engine subdir? Just make another subdir called libs.

ProjectRevoTPP:
I mean, I tried to put it in libs/ rather than src/libs.

ProjectRevoTPP:
I am debating on a menu/ folder as well. Thoughts?

PR #408, opened 2017-08-14
decompile pokeblock feed

DizzyEggg:


src/pokeblock_feed.c
```diff
@@ -0,0 +1,1012 @@
...
+struct Sprite gPokeblockFeedPokeSpriteCopy;
+u16 gUnknown_03005F94;
+s16 gUnknown_03005FA0[24];
+
+// rodata
+
+static const u8 sNatureToMonPokeblockAnim[][2] =
+{
+    {  0, 0 }, // HARDY
+	{  3, 0 }, // LONELY
```
camthesaxman: tabs

src/pokeblock_feed.c
```diff
@@ -0,0 +1,1012 @@
...
+{
+    AFFINEANIMCMD_FRAME(0, 0, 12, 1),
+    AFFINEANIMCMD_FRAME(0, 0, 0, 30),
+    AFFINEANIMCMD_FRAME(0, 0, -12, 1),
+    AFFINEANIMCMD_END
+};
+
+static const union AffineAnimCmd sSpriteAffineAnim_8411EC0[] =
+{
+    AFFINEANIMCMD_FRAME(0xFF00, 0x100, 0, 0),
```
camthesaxman: -0x100 instead of 0xFF00 would be better, since this field is signed.

src/pokeblock_feed.c
```diff
@@ -0,0 +1,1012 @@
...
+            REG_IME = savedIME;
+            REG_DISPSTAT |= 8;
+            SetVBlankCallback(VBlankCB_PokeblockFeed);
+            gMain.state++;
+        }
+    case 13:
+        BeginNormalPaletteFade(-1, 0, 0x10, 0, 0);
+        gPaletteFade.bufferTransferDisabled = 0;
+        SetMainCallback2(CB2_PokeblockFeed);
+        return 1;
```
camthesaxman: Use `TRUE` and `FALSE` since it's bool.

src/pokeblock_feed.c
```diff
@@ -0,0 +1,1012 @@
...
+                gUnknown_03005F3C = 70;
+            }
+        }
+        break;
+    case 70:
+        FreePokeSpriteMatrix();
+        gUnknown_03005F40 = 0;
+        gUnknown_03005F3C = 0;
+        break;
+    case 71 ... 90:
```
camthesaxman: Maybe try changing this to `case 90` here and getting rid of the `case 1 ... 9` above. I've encountered things like this in other files, so those intermediate cases are just skipping the switch.

src/pokeblock_feed.c
```diff
@@ -0,0 +1,1012 @@
...
+        gUnknown_03005FA0[4] = gUnknown_03005FA0[12];
+        return 0;
+    }
+}
+
+static bool8 sub_81485CC(void)
+{
+    u16 var = gUnknown_03005FA0[12] - gUnknown_03005FA0[4];
+
+    gPokeblockFeedPokeSprite->pos2.x = (*((u16*)(&ewram[0x1D000]) + var));
```
camthesaxman: Does it match if you do this?
`#define ewram1D000 ((u16 *)(ewram + 0x1D000))`
`gPokeblockFeedPokeSprite->pos2.x = ewram1D000[var];`

src/pokeblock_feed.c
DizzyEggg: Sadly, both ... are needed, otherwise it will get optimized.

src/pokeblock_feed.c
DizzyEggg: good idea, it worked quite nicely

PR #412, opened 2017-08-21
finish decompiling shop.c

ProjectRevoTPP:


src/shop.c
```diff
@@ -524,3 +531,95 @@ void sub_80B37F8(u8 taskId)
     MenuPrint(gOtherText_xString1, 1, 11);
     sub_80A3FA0(gBGTilemapBuffers[1], 1, 11, 12, 2, 0xC3E1);
 }
+
+void sub_80B389C(u16 itemId, u8 var2, bool32 hasControlCode)
+{
+    u8 *stringPtr = gStringVar1;
+
+    if(hasControlCode != FALSE)
```
camthesaxman: Please put spaces after `if`, `for`, `while`, and `switch`.

ProjectRevoTPP:
Before this is merged, I really wanna get a few, if not all, of the indented asm functions decompiled. I skipped these because they looked really really hard, 2 of which accessed gMenuWindow.tilemap in a way ive never seen other functions do.

PR #414, opened 2017-08-23
organize battle buffer files

DizzyEggg:
As the title says.


camthesaxman:
I'll merge battle_5 so you can add that also.

DizzyEggg:
I'm having way too much troubles merging it :/ Could you help me with it?

camthesaxman:
You can copy the changed files from here. https://github.com/camthesaxman/pokeruby/tree/DizzyEggg-organize_battle_buffer_files For some reason, I can't push to your branch. Also, rename the battlecmd files to battle_controller.

src/rom3.c
```diff
@@ -1060,7 +1060,7 @@ void sub_800CBE0(u8 a, u8 *b)
     dp01_prepare_buffer(a, gBattleBuffersTransferData, 4);
 }
 
-void EmitChoosePokemon(u8 a, u8 b, u8 c, u8 d, u8 *e)
+void Emitcmd22(u8 a, u8 b, u8 c, u8 d, u8 *e)
```
camthesaxman: Any reason why you changed this one? Is `EmitChoosePokemon` not correct?

DizzyEggg: I copied all the names from player buffers and applied them to all files. This one wasn't named in the player controller, so I gave it a temporary name. So yeah, you can change it back to EmitChoosePokemon.

DizzyEggg:
Cam already took care of this, so Imma close it.

PR #419, opened 2017-08-25
(don't merge yet) use macro to make enums available in both C and assembly

camthesaxman:


camthesaxman:
This is obsolete since we decided to use files in include/constants.

PR #431, opened 2017-09-10
Field effect helpers

PikalaxALT:


include/field_effect_helpers.h
```diff
@@ -1,6 +1,8 @@
 #ifndef GUARD_FIELD_EFFECT_HELPERS_H
 #define GUARD_FIELD_EFFECT_HELPERS_H
 
+extern const struct SpriteTemplate *const gFieldEffectObjectTemplatePointers[];
```
huderlem: unecessary? doesn't seem to be referenced anywhere.

src/field_effect_helpers.c
```diff
@@ -0,0 +1,1632 @@
...
+        }
+    }
+}
+
+u8 sub_8126B54(void)
+{
+    u8 spriteId;
+    struct Sprite *sprite;
+
+    spriteId = CreateSpriteAtEnd(gFieldEffectObjectTemplatePointers[8] /*gFieldEffectSpriteTemplate_Arrow*/, 0, 0, 0x52);
```
huderlem: Is this comment left over, or does it serve a purpose?

PikalaxALT: for a while during this file's decomp i was keeping track of what this access was loading. i tired of that after a few functions. eventually we want to consider making an enum for these.

PR #432, opened 2017-09-11
(don't merge yet) convert map data to C

camthesaxman:


camthesaxman:
Abandoning this for now.

PR #440, opened 2017-10-02
allow exporting PNG palettes with gbagfx

camthesaxman:


camthesaxman:
I don't know of any way to get this to work with Make, so I'm closing this. Adding a handler to convert .png to .gbapal would work better with our build system.

yenatch:
It doesn't need to be in the makefile to be worth adding to gbagfx.

PR #445, opened 2017-10-15
Cable car and Cable car util

PikalaxALT:


src/scene/cable_car.c
```diff
@@ -297,22 +297,22 @@ static void sub_8123244(void)
             REG_IME = 0;
             REG_IE |= INTR_FLAG_VBLANK;
             REG_IME = imebak;
-            SetVBlankCallback(sub_8123C40);
-            SetMainCallback2(sub_8123724);
+            SetVBlankCallback(CableCarVblankCallback);
```
ProjectRevoTPP: why not just VBlank_CB as we usually call it if it's static?

PR #448, opened 2017-10-18
 document bike.c.

ProjectRevoTPP:
The first of many PRs. I hope this one becomes the gold standard to follow.

Removed all unknown labels from bike.c. Named all in-file functions. Sub count down to 7 in bike.c.


include/bike.h
```diff
@@ -22,6 +23,15 @@ enum
     SPEED_FASTEST,
 };
 
+// mach bike transitions enum
+enum
+{
+    MBT_FACE_DIRECTION,
```
huderlem: Either change these to "MACH_TRANS" or change ACRO_TRANS below to "ABT"

huderlem:
Otherwise looks fine to me

PR #454, opened 2017-11-05
ewram refactor

ProjectRevoTPP:
This is going to be messy. I also want to update all direct ewram uses to use defines instead, but that may take a while.


ProjectRevoTPP:
I think that's actually going to be it for this PR. If I add anymore, it's going to be a conflict nightmare.

include/battle.h
```diff
@@ -234,7 +234,11 @@ struct AI_ThinkingStruct /* 0x2016800 */
 
 struct BattleStruct /* 0x2000000 */
 {
-    u8 filler0[0x15DDE];
+    /*0x00000*/ u8 unk0;
```
ProjectRevoTPP: This was part of an early attempt to merge ewram structs. This can be ignored as I don't know if its actually part of this.

ProjectRevoTPP:
Screw it, I'm going for gold. Go big or go home.

PR #461, opened 2017-11-13
migrate data

ProjectRevoTPP:


src/battle/battle_message.c
```diff
@@ -44,66 +174,64 @@ extern u16 gBattlePartyID[4];
...
-extern const u8 gUnknown_084007C8[];
-extern const u8 gUnknown_084007CA[];
-extern const u8 gUnknown_084007CC[];
-extern const u8 gUnknown_084007CE[];
-extern const u8 gUnknown_084007D0[];
-extern const u8 gUnknown_08400E5E[];
-extern const u8 gUnknown_08400E62[];
-
-extern const u16 gUnknown_084016BC[]; // a table of moves
+extern const u8 BattleText_DoubleWantToBattle[];
```
huderlem: Since you're now including ../data/battle_strings_en.h, all of these extern declarations aren't needed.

ProjectRevoTPP:
Hopefully this is enough for pokeruby to reach 50% in C.

src/field/field_door.c
```diff
@@ -6,16 +6,397 @@
...
+static const struct DoorAnimFrame gDoorCloseAnimFrames[] =
+{
+    {4, 0x200},
+    {4, 0x100},
+    {4, 0},
+    {4, -1},
+    {0, 0},
+};
+
+const u8 DoorAnimPalettes_830F8A4[] = {1, 1, 1, 1, 1, 1, 1, 1}; // door 00
```
huderlem: Are these palettes or palette ids?  If they are palette ids, that should be reflected in their name to avoid confusion.

src/field/tv.c
```diff
@@ -105,6 +86,351 @@ static u8 gUnknown_03000720;
 static u8 gUnknown_03000721;
 static s8 gUnknown_03000722;
 
+const struct OutbreakPokemon gPokeOutbreakSpeciesList[5] =
```
huderlem: I believe we want to prefix static rodata with "s", rather than "g".  We should address that later, though.

PR #463, opened 2017-11-19
decompile fldeff_recordmixing

sceptillion:


src/field/fldeff_recordmixing.c
```diff
@@ -0,0 +1,82 @@
...
+        sprite->pos1.x += 16;
+        sprite->pos1.y += 2;
+    }
+
+    return spriteId;
+}
+
+void DestroyRecordMixingSprite(void)
+{
+    int i;
```
huderlem: `u32` instead of `int`

yenatch: how come?

huderlem: We discussed this, and I was mistaken.  Using `int` is preferred when the size of the var doesn't matter.  This will commonly be the case for loops counters.

src/field/fldeff_decoration.c
```diff
@@ -0,0 +1,309 @@
...
+    s16 x, y;
+
+    ScriptContext2_Enable();
+    GetXYCoordsOneStepInFrontOfPlayer(&x, &y);
+    gFieldEffectArguments[5] = x;
+    gFieldEffectArguments[6] = y;
+
+    switch (player_get_direction_lower_nybble())
+    {
+    case 1:
```
huderlem: The `DIR_` enum values should be used for these cases.
```
enum
{
    DIR_NONE,
    DIR_SOUTH,
    DIR_NORTH,
    DIR_WEST,
    DIR_EAST,
};
```

PR #464, opened 2017-11-20
label battle anims

sceptillion:


data/battle_anim_scripts.s
```diff
@@ -382,40 +382,40 @@ gBattleAnims_StatusConditions:: @ 81C76F8
...
-	.4byte Unknown_81D6821
-	.4byte Unknown_81D6852
-	.4byte Unknown_81D6877
-	.4byte Unknown_81D68D5
-	.4byte Unknown_81D6934
-	.4byte Unknown_81D69F9
-	.4byte Unknown_81D6A39
-	.4byte Unknown_81D6A7C
+gBattleAnims_General:: @ 81C771C
+	.4byte General_CastformChange
```
huderlem: I feel these should have a "battle animation" prefix, rather than just "General" or "Special".  Maybe "BattleAnimGeneral".

sceptillion: Why? They are local labels, so they only exist in this file.

PR #465, opened 2017-11-20
clean up unknown_task variables

sceptillion:
gUnknown_03004DC0 and gUnknown_03004DE0 had a few different type declarations in various places. I unified them all to the same one, which appears to be the correct one.

### DIFF
--- a/src/battle/battle_2.c
+++ b/src/battle/battle_2.c
@@ -31,13 +31,6 @@
 #include "battle_move_effects.h"
 #include "ewram.h"
 
-struct UnknownStruct6
-{
-    u16 unk0[0xA0];
-    u8 fillerA0[0x640];
-    u16 unk780[0xA0];
-};
-
 struct UnknownStruct7
 {
     u8 unk0;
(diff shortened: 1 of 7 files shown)


src/battle/battle_2.c
```diff
@@ -1153,14 +1144,14 @@ void c2_8011A1C(void)
...
     for (i = 0; i < 80; i++)
     {
-        gUnknown_03004DE0.unk0[i] = 0xF0;
-        gUnknown_03004DE0.unk780[i] = 0xF0;
+        gUnknown_03004DE0[0][i] = 0xF0;
+        gUnknown_03004DE0[1][i] = 0xF0;
     }
     for (i = 80; i < 160; i++)
     {
         asm(""::"r"(i));  // Needed to stop the compiler from optimizing out the loop counter
```
camthesaxman: Is this asm hack still needed? Sometimes, I have to do stuff like this when I don't have the format quite right.

sceptillion: Yes. I tried to fix it, but I still can't get it to match without that.

PR #466, opened 2017-11-20
Vendor data scripts in linguist.

ProjectRevoTPP:


ProjectRevoTPP:
This is quite a big change in terms of linguist percentages. A majority of the remaining asm count in github comes from these scripts: once these scripts are treated as Text it will become apparent what the actual remaining assembly count is and how close we are to 100% C in terms of portable data files and undecompiled files.

ProjectRevoTPP:
Is it better to Text the scripts or vendor them completely? I need your input.

If we Text them, then its a little more representative of what the scripts actually are, but the repository will probably show up as a "Text" based one, which is wrong.
If we vendor them, then the repository will show up as a C based one, but it means jettisoning out the scripts completely from the equation, which is unfair.

This problem is really due to linguist being so inflexible.

PR #473, opened 2017-11-25
split battle_anim_80CA710

ProjectRevoTPP:
There's much more work to be done. I've split the decompiled chunk of 80C, but there's still the matter of the remaining assembly.

The files seem to largely be in a pattern, depending on the data declarations. It always seems to declare anims, anim tables, then the sprite templates, sometimes the templates are in groups before proceeding to the next file, unless that template contains a dummy. If that's the case, it could be its own file, depending on if it has anything to do with the previous one.

This is a start, but there's a lot of garbage.


ProjectRevoTPP:
Since this is a huge split, I need some comments on the boundries.

In the later assembly, it seems like GF grouped together types as a single file. Figuring out the individual general actions is too tough: it may have been by folder, like anim/ice/foo.c.

huderlem:
Same feedback as before; I think it makes more sense to have src/battle/anim.  padz suggested src/battle/anim/sprite/ for the indidividual effects.  Not sure if there would be things under both src/battle/anim/... and src/battle/anim/sprite/...

Makefile
```diff
@@ -40,9 +40,9 @@ VERSIONS := ruby sapphire ruby_rev1 sapphire_rev1 ruby_rev2 sapphire_rev2 ruby_d
 $(VERSIONS) $(VERSIONS:%=compare_%)
 
 
-$(shell mkdir -p build/ $(VERSIONS:%=build/%/{,asm,data,src{,/battle,/field,/debug,/scene,/pokemon,/engine,/libs,/anim}}))
+$(shell mkdir -p build/ $(VERSIONS:%=build/%/{,asm,data,src{,/battle{,/anim},/field,/debug,/scene,/pokemon,/engine,/libs,/anim}}))
```
huderlem: Don't need this "/anim" anymore.

PR #492, opened 2017-12-17
Decompile data and abstract away some explicit memcpy calls

PikalaxALT:


ld_script.txt
```diff
@@ -600,7 +600,7 @@ SECTIONS {
         src/debug/unknown_debug_menu.o(.rodata);
         data/strings2.o(.rodata);
         src/engine/menu_cursor.o(.rodata);
-        data/mystery_event_msg.o(.rodata);
+        src/engine/mystery_event_script.o(.rodata);
```
sceptillion: This causes the order of files to differ between the .text and .rodata sections. mystery_event_msg is probably its own file, not part of mystery_event_script.

scnorton: How important is preserving file order? There appear to be other instances where file order matching is violated 🤔 💭

sceptillion: Well, it's only important if we want this source to resemble the original source code. The other violations should be fixed as well, if possible.

src/engine/naming_screen.c
```diff
@@ -2030,4 +2029,162 @@ const struct OamData gOamData_83CE4A8 =
...
+    11, 3, &gOamData_83CE498, gSpriteAnimTable_83CE5B8, NULL, gDummySpriteAffineAnimTable, sub_80B6DE8
+};
+
+const struct SpriteTemplate gSpriteTemplate_83CE688 = {
+    0xFFFF, 0, &gOamData_83CE498, gSpriteAnimTable_83CE5C4, gSpriteImageTable_83CE588, gDummySpriteAffineAnimTable, SpriteCallbackDummy
+};
+
+const struct SpriteSheet gUnknown_083CE6A0[] = {
+    { gNamingScreenBackButtonTiles, 0x1E0, 0 },
+    { gNamingScreenOKButtonTiles, 0x1E0, 1 },
```
ProjectRevoTPP: OK

src/field/decoration.c
```diff
@@ -1605,8 +1472,7 @@ void DoPlayerPCDecoration(u8 taskId)
 
 void sub_80FE2B4(void)
 {
-    u16 palettes[3];
-    memcpy(palettes, gUnknown_083EC654, sizeof gUnknown_083EC654);
+    u16 palettes[3] = {0x6318, 0x739C, 0x7FFF};
```
ProjectRevoTPP: use the RGB macro if possible

huderlem:
Entries in bard_music_en.h correspond directly to easy chat words, correct?  I think it would be helpful if the easy chat words were in comments next their respective entries.  Also, the entries could go on single lines, so that the file isn't so long (currently 14k lines).

src/libs/m4a_2.c
```diff
@@ -85,7 +85,8 @@ void m4aSoundInit(void)
         mplayInfo->memAccArea = gMPlayMemAccArea;
     }
 
-    memcpy(&gPokemonCrySong, &gPokemonCrySongTemplate, sizeof(struct PokemonCrySong));
+    gPokemonCrySong = gPokemonCrySongTemplate;
+//    memcpy(&gPokemonCrySong, &gPokemonCrySongTemplate, sizeof(struct PokemonCrySong));
```
huderlem: Any reason to keep this comment?

src/pokemon/pokemon_2.c
```diff
@@ -1060,7 +1060,7 @@ void CreateSecretBaseEnemyParty(struct SecretBaseRecord *secretBaseRecord)
     s32 i, j;
 
     ZeroEnemyPartyMons();
-    memcpy(eSecretBaseRecord, secretBaseRecord, sizeof(*secretBaseRecord));
+    *eSecretBaseRecord = *secretBaseRecord;
```
huderlem: I like that all these memcpys are disappearing.

src/libs/m4a_2.c
PikalaxALT: No, I just forgot to delete it.  I left it in case my change would have broken it

PR #495, opened 2017-12-20
Add NDEBUG support to pokeruby.

ProjectRevoTPP:
This PR is not going to pass until agbcc starts including the libc.a changes and pokeemerald updates as well. Travis won't be building this until this change is made to agbcc using sceptillion's changes for FR/LG.


ProjectRevoTPP:
As of recent, Ive been told by endrift that mGBA has working AGBPrint support (most likely in the next release whenever that is). So not only will VBA work but mGBA as well, though i would still like to hack in no$GBA print support somehow.

ProjectRevoTPP:
Ive submitted NDEBUG PRs to all 3 repositories as of now.

Makefile
```diff
@@ -16,6 +16,8 @@ OBJCOPY := $(DEVKITARM)/bin/arm-none-eabi-objcopy
 LIBGCC := tools/agbcc/lib/libgcc.a
 LIBC   := tools/agbcc/lib/libc.a
 
+LIBC := tools/agbcc/lib/libc.a
```
huderlem: This line is no longer necessary. (It's' repeated two lines above)

ProjectRevoTPP: merge must have added this. whoops.

include/config.h
```diff
@@ -1,6 +1,21 @@
 #ifndef GUARD_CONFIG_H
 #define GUARD_CONFIG_H
 
+// In the Generation 3 games, Asserts were used in various debug builds.
```
huderlem: We should add a short "call to action" like this to the beginning of this comment block.

"To enable print debugging, uncomment "#define NDEBUG".  This allows the various AGBPrint functions to be used. (See include/gba/isagbprint.h).  Some emulators support a debug console window. (And maybe call out no$gba's special function)

ProjectRevoTPP:
This PR seems complete enough. I just need it to be merged along with the other PRs in the other repositories.

PR #509, opened 2018-01-05
Decompile various battle_anim files

progreon:


src/battle/anim/wisp_fire.c
```diff
@@ -0,0 +1,48 @@
...
+
+    sprite->data[3] += 0xC0 * 2;
+    sprite->data[4] += 0xA0;
+
+    sprite->pos2.x = Sin(sprite->data[1], sprite->data[3] >> 8);
+    sprite->pos2.y = Cos(sprite->data[1], sprite->data[4] >> 8);
+
+    sprite->data[1] = (sprite->data[1] + 7) & 0xFF;
+
+    if (gMain.inBattle)
```
camthesaxman: Please put braces here. You have a dangling else.

src/battle/anim/wisp_fire.c
```diff
@@ -0,0 +1,48 @@
...
+        if ((u16) (sprite->data[1] - 0x40) > 0x83)
```
camthesaxman: This is equivalent to `if (sprite->data[1] < 64 || sprite->data[1] > 195)`.

progreon: Oh wow, totally didn't think about that, thanks!

src/battle/anim/bug.c
```diff
@@ -0,0 +1,105 @@
...
+    sprite->data[5] = gBattleAnimArgs[3];
+    sprite->callback = sub_80DCA38;
+}
+
+static void sub_80DCA38(struct Sprite *sprite)
+{
+    if (sub_8078B5C(sprite))
+    {
+        DestroyAnimSprite(sprite);
+        return;
```
progreon: Is this return fine, or do you rather have an if-else?
Destroy + return has to be used in another function of another file so I continued using it like that without realising.

asm/bug.s
```diff
@@ -7,285 +7,285 @@
...
+# 	ldrh r0, [r4, 0x3A]
+# 	adds r0, 0xD
+# 	movs r1, 0xFF
+# 	ands r0, r1
+# 	strh r0, [r4, 0x3A]
+# _080DCA6A:
+# 	pop {r4}
+# 	pop {r0}
+# 	bx r0
+# 	thumb_func_end sub_80DCA38
```
ProjectRevoTPP: please remove after you decompile

progreon:
This may be reviewed and merged for now, will continue this in a few days.

PR #512, opened 2018-01-06
[Merge if you so choose] Pokemon Storage System

PikalaxALT:


src/pokemon/pokemon_storage_system.c
```diff
@@ -1,93 +1,711 @@
...
+}
+
+#ifdef NONMATCHING
+void unref_sub_8095D08(u16 *dest, u16 dest_left, u16 dest_top, u16 width, u16 height)
+{
+    u16 i;
+
+    dest += dest_top * 0x20 + dest_left;
+    width *= 2;
+    for (i = 0; i < height; dest += 0x20, i++)
```
ProjectRevoTPP: This looks similar to the type of modified DmaLoop that ProcessDma3Requests uses where it checks size (first) before the loop. Have you tried using that macro instead?

include/pc_screen_effect.h
```diff
@@ -0,0 +1,21 @@
+#ifndef POKERUBY_PC_SCREEN_EFFECT_H
```
camthesaxman: `GUARD_PC_SCREEN_EFFECT_H`

PR #513, opened 2018-01-07
Move animation sound stuff

progreon:
If someone could help me with that nonmatching function, that'd be great! It's very close but I just don't seem to get it...


progreon:
Similar issue with the next nonmatching

progreon:
Finished decompiling this, but I wonder if this file should be split. At least I don't think sfx is a good name for this.

PR #531, opened 2018-01-14
Decompiling battle_anim data

progreon:


progreon:
All the battle_anim data I'll do for now (can be reviewed & merged)

src/battle/battle_anim.c
```diff
@@ -13,13 +13,1855 @@
...
 #include "ewram.h"
+#include "graphics.h"
 
 // sprites start at 10000 and thus must be subtracted of 10000 to account for the true index.
 #define GET_TRUE_SPRITE_INDEX(i) (i - 10000)
 
 #define ANIM_SPRITE_INDEX_COUNT 8
 #define ANIM_ARGS_COUNT 8
 
+const struct OamData gOamData_837DF24 =
```
huderlem: Maybe all of this .rodata should get its own `src/data/battle_anim.h` file?

camthesaxman: The file is still under 5000 lines, so I think this is okay.

huderlem: I guess it's fine for now.  Having to scroll down 2k lines to reach the first function is a bit cumbersome, though.

progreon: Yeah I had it in a separate header file first, but when I asked about it cam said to just put it in there, which is why I did

PR #535, opened 2018-01-14
fixes

ProjectRevoTPP:


ProjectRevoTPP:
This is enough.

include/shop.h
```diff
@@ -17,6 +17,14 @@ enum
     ANIM_NUM
 };
 
+// shop options
+enum
+{
+    BUY,
```
huderlem: These should probably be prefixed with `SHOP_`

include/menu_helpers.h
```diff
@@ -9,7 +9,22 @@ struct YesNoFuncTable
...
+{
+    TOP_ARROW,
+    BOTTOM_ARROW,
+    LEFT_ARROW, // Bag arrows
+    RIGHT_ARROW
+};
+
+enum
+{
+    VISIBLE,
```
huderlem: This could use a prefix, too. Perhaps `SCROLL_INDICATOR_`

PR #540, opened 2018-01-16
port some fixed battle code from pokeemerald

camthesaxman:


camthesaxman:
I will continue porting code after I clean things up in both pokeruby and pokeemerald.

PR #541, opened 2018-01-17
build German debug version

camthesaxman:


camthesaxman:
TODO: remove the raw jump addresses in event_scripts.s

undefined.txt
```diff
@@ -0,0 +1,29 @@
+(.text+0x120): undefined reference to `gUnknown_02023A14_50'
```
huderlem: This file appears to be committed by mistake.

huderlem:
Is there an alternative to directly including the assembly?  I think it's making the C files too unwieldy in many cases, and who knows when they would eventually get decompiled.

camthesaxman:
I can move the undecompiled code to asm, if that's better.

camthesaxman:
I decided to leave the code in start_menu_debug.c since I already started decompiling it and added the correct prototypes for most of the functions.

camthesaxman:
For those files that have a lot of undecompiled debug code, would it be a good idea to split those into .h files for the time being and `#include` them? I don't like having that huge mess of asm in a file, but at the same time, I don't want to "swiss cheese" them, especially when the debug code is in the middle of the file, like battle_2.c.

PR #547, opened 2018-01-21
Use standard flags to link libgcc, libc

PikalaxALT:


yenatch:
pret/pokeemerald#178 breaks windows builds

camthesaxman:
I don't really see the point of this, if you have to link the archives directly on Windows. It's better to be consistent and do the same thing on Linux, OS X, and Windows.

PikalaxALT:
@camthesaxman The hope is that the devkitARM team will resolve the bug that prevents traditional library linking in cygwin.

PikalaxALT:
This PR is obsolete.

PR #549, opened 2018-01-21
Created a git hook to warn the user if they messed up.

easyaspi314:
This will prevent the user from making files executable or not running
./de_after.sh.

Currently, the way we install the hook is in the Makefile in a rather dirty
way. It bugs me that pre-commit hooks can't be specified beforehand.

Basically, I put an empty file .GERMAN in `data-de/` and `graphics-de/`. These will be copied to `data` and `graphics` when you run `./de_before.sh`. If the user doesn't run `./de_after.sh`, `data/.GERMAN` and `graphics/.GERMAN` will exist and the pre-commit hook will fail.

Additionally, I check any added files for the executable permission aside from `.hooks/pre-commit`, and any shell scripts.

### DIFF
--- a/Makefile~
+++ b/Makefile~
@@ -1,0 +1,174 @@
+
+include config.mk
+
+
+#### Tools ####
+
+SHELL     := /bin/bash -o pipefail
+AS        := $(DEVKITARM)/bin/arm-none-eabi-as
+CC1       := tools/agbcc/bin/agbcc
+CPP       := $(DEVKITARM)/bin/arm-none-eabi-cpp
+LD        := $(DEVKITARM)/bin/arm-none-eabi-ld
+OBJCOPY   := $(DEVKITARM)/bin/arm-none-eabi-objcopy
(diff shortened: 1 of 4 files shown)


Makefile~
```diff
@@ -0,0 +1,174 @@
+
```
yenatch: this file should probably not be here

easyaspi314:
Oh wow

easyaspi314:
Remember when I said "wildcards are evil?"

yenatch:
It would be much better to fix the german build so it can be run along the others rather than bandaiding it.

A hook for executable files is maybe something we could ask contributors to use, but it would just punish anyone using the repo on their own.

easyaspi314:
Hmm. I wonder if it could be tested in Travis?

yenatch:
Yes. That would be good.

easyaspi314:
Then screw this.

PR #551, opened 2018-01-22
Contest link 80 c857 c

PikalaxALT:


src/battle/contest_link_80C857C.c
```diff
@@ -1,9 +1,732 @@
...
+    }
+}
+
+void sub_80C8E1C(u8 taskId)
+{
+    int i;
+
+    switch (gTasks[taskId].data[0])
+    {
+#if ENGLISH
```
huderlem: I wonder if it's better to surround the whole functions with the `#if` blocks.  Having them inline everywhere makes it illegible.

yenatch: at the very least this:
```c
#if GERMAN
                    if (gTasks[taskId].data[0] == 0)
                    {
                        gTasks[taskId].data[0] = 3;
                    }
                    else
                    {
                        de_sub_80C9274(FALSE);
#endif
                    sub_8007E9C(2);
#if ENGLISH
                    gTasks[taskId].data[0]++;
#elif GERMAN
                    gTasks[taskId].data[0] = 1;
                    }
#endif
```

```c
#if GERMAN
                    if (gTasks[taskId].data[0] == 0)
                    {
                        gTasks[taskId].data[0] = 3;
                    }
                    else
                    {
                        de_sub_80C9274(FALSE);
                        sub_8007E9C(2);
                        gTasks[taskId].data[0] = 1;
                    }
#else
                    sub_8007E9C(2);
                    gTasks[taskId].data[0]++;

#endif
```

aaaaaa123456789: Agreed. Blocks shouldn't cross `#if`/`#endif` boundaries.

PR #553, opened 2018-01-25
merge field_map_obj helpers and ground effect into event_object_movement.c

ProjectRevoTPP:
The Fire Red asserts tells us these should be the same file. So, I combined them in this PR for accuracy. When emerald has the same files decompiled, these will be merged too.


huderlem:
I can't tell what those letters are supposed to stand for.  Probably needs a friendlier name.

ProjectRevoTPP:
evobjmv is what it was called in the assert.

huderlem:
Sure, but we can do better.  `event_object_move.c` is an improvement.  It's impossible to know what `evobjmv.c` means without going through the file and reading it.

PR #560, opened 2018-02-02
IDE Integrations (cMake, Programmer's Notepad, Sublime Text)

ProjectRevoTPP:
This is an ongoing WIP effort. Eventually, stuff like CMake or Programmers Notepad will go in projects/ as user friendly integrations for pokeruby making it much easier to get setup with a developer environment.

For Visual Studio, I'm still reorganizing the filters. I moved all map scripts to "Scripts", but I don't know if that  should be the only change.


ProjectRevoTPP:
I plan to add Programmer's Notepad and CLion integrations to this PR as well, so I am going to need some guidance and advice on how to approach this problem.

projects/pokeruby.sublime-project
```diff
@@ -0,0 +1,8 @@
+{
+	"folders":
+	[
+		{
+			"path": "..\\"
```
camthesaxman: Please make this `"../"`. Backslashes are Windows only.

ProjectRevoTPP:
If I want proper VS support, I think I would need a full cmake project so the project can be generated. The other integrations here are reasonable.

PR #579, opened 2018-03-04
Use cygwin on windows and build all the tools automatically.

yenatch:


PikalaxALT:
```
make -j`nproc`
```
will use all available CPU power to build the ROM.  Not recommended if you plan to do something else while compiling in the background.  Perhaps
```
make -j$(($(nproc)/2))
```?

PR #596, opened 2018-04-17
Watanabe

PikalaxALT:


Makefile
```diff
@@ -1,23 +1,27 @@
 
 include config.mk
 
+ifeq ($(OS),Windows_NT)
```
huderlem: Consider explicitly setting `EXE` to empty in `else` block, just in case `EXE` is already defined in the env.

PR #599, opened 2018-04-22
[WIP] battle/anim/water.c

PikalaxALT:


Doesnty:
So there's other more confusing parts to sub_80D38BC not matching, but I took a look for selfish reasons at getting it to be function and all that seems to be needed is `gTasks[taskId].data[0x6] = 1;` before the last line of the function.

PikalaxALT:
Superseded by #716

PR #615, opened 2018-05-17
Rename sub_804C1A8() in trade.c to SetTradeSceneStrings()

Deokishisu:
sub_804C1A8() is responsible for getting the variable strings (like the trade partner's OTName, their Pokemon's nickname, and your Pokemon's nickname) needed to print the messages in the trade scene, so I renamed it to SetTradeSceneStrings(). Added a comment in the function to describe what it's doing as well.

Please let me know if this name isn't accurate enough or if my comment should be in a different style.


huderlem:
I think 'SetTradeSceneStrings' is better. 'Get' implies that the function returns something.

PR #628, opened 2018-06-14
updates to better match Emerald

Mother-Of-Dragons:


src/tileset_anim.c
```diff
@@ -237,62 +237,62 @@ static const void *const gUnknown_0837BB04[] =
...
+    gTilesetAnims_Mauville0a_Frame3,
+    gTilesetAnims_Mauville0a_Frame3,
+    gTilesetAnims_Mauville0a_Frame3,
+    gTilesetAnims_Mauville0a_Frame3,
+    gTilesetAnims_Mauville0a_Frame3,
+    gTilesetAnims_Mauville0a_Frame2,
+    gTilesetAnims_Mauville0a_Frame1,
 };
 
 static const u8 *const gTilesetAnimTable_Mauville_1A[] =
```
huderlem: I"m noticing that the names of a few of these table aren't matching the names of the entries in the table. For example, this one has `Mauville_1A` in its name, but the entries have `Mauville0b`.  Are they supposed to match?

PR #646, opened 2018-06-29
libisagbprn added volatile qualifier to touched pointers

felixjones:
This fixes libisagbprn as a standalone library for compilers that eagerly optimise away pointer access.


huderlem:
This does not match in pokefirered, so closing, as per https://github.com/pret/pokeemerald/pull/271

PR #652, opened 2018-07-08
[WIP] Decompile ice.s

garakmon:
so far:
11 MATCHING
2 NONMATCHING


src/battle/anim/ice.c
```diff
@@ -1220,11 +1044,96 @@ void sub_80D7E88(struct Sprite *sprite)
...
+    if (gBattleAnimArgs[4] == 0)
+    {
+        if (gBattleAnimArgs[5] == 0)
+            InitAnimSpritePos(sprite, 0);
+
+        else
+        {
+            SetAverageBattlerPositions(gAnimBankAttacker, 0, &sprite->pos1.x, &sprite->pos1.y);
+
+            if (GetBattlerSide(gAnimBankAttacker) << 24 != B_SIDE_PLAYER)
```
huderlem: The shift is not needed

ProjectRevoTPP:
this can be merged now

garakmon:
Ok I labeled the functions as well as I could. Does somebody wanna take a (quick?) look over them? Specifically the ones with the words "Ice" or "Snow" in them. I wasn't sure how descriptive to get with the names, or whether more general names would be better (eg. `AnimMoveParticleTowardsTarget` instead of `AnimIceBeamParticle`) because the functions aren't necessarily move-specific.

PR #658, opened 2018-07-21
Copy the Trainer struct from pokeemerald and regen the trainers.h files.

vblm12:
I also renamed the OPPONENT constants to TRAINER to better match pokeemerald.


garakmon:
It looks like you missed a few OPPONENT_NAME to TRAINER_NAME switches in src/debug/taya_debug_window.c

PR #662, opened 2018-07-27
Improve build process with better tools.

easyaspi314:
- Instead of printing dependencies to stdout, scaninc makes a Makefile dependency file like gcc does when you run `gcc -MT $@ -MMD -MP -MF $(DEPDIR)/$*.Td`. Because of this, scaninc no longer runs unnecessarily:
  - These dependencies are stored on disk, so we don't have to keep scanning each time we run `make`.
  - These dependency files are seen as dependencies when `NODEP` is not set, and will automatically be updated by `make` when needed.
- scaninc is a little more efficient, ~~but it needs C++14, which is pretty standard now.~~


mid-kid:
I've seen this technique before, but it relied on re-running make after the dependencies were built, because make doesn't start doing anything until the entire dependency tree is known, so it won't build the dependency files before including them.
That's probably the reason why Travis failed.

easyaspi314:
@mid-kid the issue is simpler:
```

g++: error: unrecognized command line option ‘-std=c++14’
```

easyaspi314:
Huh. I need to do some more testing. Build passes, but make gives no such file or directory messages.

huderlem:
Haven't been able to test this out yet. Does @mid-kid's point about removal of graphics apply to these changes?  If so, does it also apply to removing non-graphics dependencies?  That would cause a lot of confusion.

Additionally, adding the ability to configure the scaninc output directory should be completed before this goes in.

easyaspi314:
```
>The stuff about removal of graphics or other dependencies still stands
```
Except it doesn't, I have that covered.

There are two parts to the dependency file. The dependencies, which states what files depend on what:
```makefile
file.o: file.c dependency.4bpp
```
…and the phonies, which make a dummy rule in case the file is deleted:
```
dependency.4bpp:
```
In make, rules only add on to other rules. In this case, what it does is make a phony target for that file if it doesn't exist (and it will just think that there is nothing to do for it), or adds nothing to the file's rules if it does exist.

As for the output filename, I can easily add that in.

PikalaxALT:
Is this still alive?

mid-kid:
If it is, at least merge https://github.com/pret/pokeruby/pull/662/commits/c3db15a1a14160c9e1281e6643060c79030d5d3c or a variation of it, even if not merging all of it.

PikalaxALT:
Closing this PR because
- The maintainer of this branch is missing in action
- The branch requires agbcc other than master. Yeah, I know, hasn't stopped me on FireRed, but Ruby gets pret/agbcc grandfathered in.

PR #671, opened 2018-08-15
Use constants defined in global.h, and add constants to match pokeemerald

hiddenruby:


include/global.h
```diff
@@ -96,7 +96,8 @@ enum LanguageId
 // capacities of various saveblock objects
 #define DAYCARE_MON_COUNT   2
 #define POKEBLOCKS_COUNT    40
-#define EVENT_OBJECTS_COUNT   16
+#define PARTY_SIZE          6
```
huderlem: `PARTY_SIZE` is already defined in `pokemon.h`--it should be removed from there.

hiddenruby: Ok. I've also since added TV_SHOWS_COUNT & POKE_NEWS_COUNT and I'm now going through replacing those so I'll change this to a WIP for now and add that when I think I'm done.

PR #679, opened 2018-08-22
 Changed pokedex_orders.h to use defines from species.h

ultima-soul:
Just as title says. Please suggest any improvements if necessary.


src/data/pokedex_orders.h
```diff
@@ -1,1609 +1,1609 @@
...
+	NATIONAL_DEX_OLD_UNOWN_R,
+	NATIONAL_DEX_OLD_UNOWN_S,
+	NATIONAL_DEX_OLD_UNOWN_T,
+	NATIONAL_DEX_OLD_UNOWN_U,
+	NATIONAL_DEX_OLD_UNOWN_V,
+	NATIONAL_DEX_OLD_UNOWN_W,
+	NATIONAL_DEX_OLD_UNOWN_X,
+	NATIONAL_DEX_OLD_UNOWN_Y,
+	NATIONAL_DEX_OLD_UNOWN_Z,
+	NATIONAL_DEX_ABRA,  // Abra
```
huderlem: all of the comments that specify a pokemon can be removed now, since the constant makes that clear

src/data/pokedex_orders.h
```diff
@@ -1,1609 +1,1609 @@
...
-    306,  // Aggron
-    321,  // Wailord
-    208,  // Steelix
-    143,  // Snorlax
-    376,  // Metagross
-    383,  // Groudon
+	NATIONAL_DEX_GASTLY,  // Gastly
+	NATIONAL_DEX_HAUNTER,  // Haunter
+    NATIONAL_DEX_HOPPIP,  // Hoppip
+	NATIONAL_DEX_DIGLETT,  // Diglett
```
huderlem: there is a mix of tabs and 4 spaces throughout this entire file--use 4 spaces for all indentation.

src/data/pokedex_orders.h
ultima-soul: Should I leave the comments for the German names?

huderlem: no, those should also be removed.

huderlem: Oh, my mistake.  I think you're right--leave the german names.

PR #680, opened 2018-08-23
Add Defines for Pokedex Related Numbers and Fixed Comment in birch_pc.c

ultima-soul:
Any suggestions are welcome.


src/birch_pc.c
```diff
@@ -104,7 +104,7 @@ const u8 *GetPokedexRatingText(u16 count)
         return gBirchDexRatingText_DexCompleted;
     }
     if (count == 202)
```
Normmatt: shouldn't this also use the HOENN_DEX_COUNT define?

ultima-soul: It can, but as you can see it's based off of the tiers of Pokedex completion after Deoxys and Jirachi checks. I figured people would want it to be customizable rather than strictly enforced so I left this one alone as they may want to remove the Jirachi and Deoxys checks and do some other custom stuff.

A different solution would be to indeed add HOENN_DEX_COUNT and then modify the previous number checks so 200 would be HOENN_DEX_COUNT - 2 and 190 being HOENN_DEX_COUNT - 12.

PR #681, opened 2018-08-27
Flying

who-knows-who:
So far:
matching - 27
fakematching - 1
nonmatching - 5

sub_80DA6F0 really needs to be looked at by someone with more experience/brain power


ProjectRevoTPP:
can we merge this already

who-knows-who:
Tidied it up a bit, don't think I can match any more, as long as people are happy with the state of the code I don't see why not?

PR #693, opened 2018-09-29
Add support for gbafix

### LINKED ISSUE #568: Checksum in the header shouldn't be hardcoded
Currently the ROM's checksum in the header is hardcoded.  `gbafix` should be called in the Makefile to dynamically calculate this value.

aplumafreak500:
Fixes #568.


src/crt0.s
```diff
@@ -13,7 +13,60 @@
 Start: @ 8000000
 	b Init
 
-	.include "asm/rom_header.inc"
+	.global RomHeaderNintendoLogo
+RomHeaderNintendoLogo:
```
huderlem: logo should no longer be included.  it's set by gbafix

PR #705, opened 2018-11-12
Rename 8083C50 to WaitForRemoteLinkPlayers

LordRyst:
Name seems to fit functionality.


LordRyst:
In my excitement, forgot to check compile and missed a definition header.

PR #711, opened 2018-12-07
Use real apostrophes instead of fake ones

melthelesbian:


ProjectRevoTPP:
unknown github PR (U+27)

PR #714, opened 2018-12-15
Document Pokémon item effects, other minor changes.

luckytyphlosion:
- Add optional RTLFLAGS variable in Makefile
- Use MEMORY in linkerscript, also pad elf file
- Add replace.sh


huderlem:
emerald has a matching ItemEffects function, so it should be ported from there.

PikalaxALT:
This can probably be closed.

ld_script.txt
```diff
@@ -3,18 +3,21 @@ ENTRY(Start)
 gNumMusicPlayers = 4;
 gMaxLines = 60;
 
+MEMORY
```
PikalaxALT: This could probably be its own PR

ld_script.txt
```diff
@@ -1226,4 +1224,18 @@ SECTIONS {
     {
         *(*);
     }
+    
+    .fillrom :
```
PikalaxALT: No need for this if you use objcopy with --gap-fill 0xFF

PR #717, opened 2018-12-22
Defined OLD_UNOWN base stats in a macro

ShinyDragonHunter:
I did this exact same thing for pokeemerald and to keep the repos in sync, I thought it'd make sense to do it here too.


ShinyDragonHunter:
i certainly can.

PR #720, opened 2018-12-30
index gRegionMapEntries by mapsec constant

garakmon:
this was the best solution I could come up with (thank you German build) but I need this to be consistent between the gen 3 repos


garakmon:
#721

PR #721, opened 2019-01-16
sync region map data

garakmon:


garakmon:
probably the only major change here to pokeruby is reverting the src/data/region_map_layout.h file back into a binary file.

PR #722, opened 2019-01-28
Use macros for EventObjectGraphicsInfo

Slawter666:


huderlem:
Shelving--same reason as https://github.com/pret/pokeemerald/pull/501

PR #728, opened 2019-03-19
itemsjson

PikalaxALT:
- Removes src/data/item{description,}s_{en,de}.h
- Adds items.json with language information for both English and German builds
- Creates itemsjson, which builds src/data/items.h and src/data/item_descriptions.h
- Edits src/item.c to include the two aforementioned files instead of the language-specific ones
- Adds Makefile rules to properly generate targets


PikalaxALT:
Implemented [jsonproc](/pret/pokeemerald/tree/jsonproc) in place of itemsjson and json2s.  Restored mapjson for the time being.

huderlem:
Going back to the JSONification philosophy--data should be converted to JSON if there is significant benefit to doing so.  Ease of tooling is probably the main determining factor for that (e.g. map editor).  Is there a strong reason for moving items data?

PikalaxALT:
Yes: Ease of incorporating other release languages.

PR #736, opened 2019-04-14
remove some statics from src/data files

garakmon:


huderlem:
Shelving--same reason as https://github.com/huderlem/porymap/pull/121

PR #742, opened 2019-07-01
Modern gcc

PikalaxALT:
Allow Pokemon Ruby to compile properly with gcc-8.3.0 and gcc-9.1.0 (devkitARM r52 and r53, respectively).

TODO:
- [x] Support for custom toolchains
- [x] Support for toolchain-supplied libc and libgcc
- [x] Random section link order
- [x] Fix crash on battle start


PikalaxALT:
To investigate: ROM crashes on battle start

PikalaxALT:
Note: Travis-CI failure for 0f19a99 is the result of backlog on their end.

PR #743, opened 2019-07-18
Name Metatiles

garakmon:


src/braille_puzzles.c
```diff
@@ -131,12 +132,12 @@ void UseFlyAncientTomb_Callback(void)
...
-    MapGridSetMetatileIdAt(14, 26, 554);
-    MapGridSetMetatileIdAt(15, 26, 555);
-    MapGridSetMetatileIdAt(16, 26, 556);
-    MapGridSetMetatileIdAt(14, 27, 3634);
-    MapGridSetMetatileIdAt(15, 27, 563);
-    MapGridSetMetatileIdAt(16, 27, 3636);
+    MapGridSetMetatileIdAt(14, 26, METATILE_ID(Cave, SealedChamberEntrance_TopLeft));
+    MapGridSetMetatileIdAt(15, 26, METATILE_ID(Cave, SealedChamberEntrance_TopMid));
+    MapGridSetMetatileIdAt(16, 26, METATILE_ID(Cave, SealedChamberEntrance_TopRight));
+    MapGridSetMetatileIdAt(14, 27, METATILE_ID(Cave, SealedChamberEntrance_BottomLeft) | METATILE_COLLISION_MASK);
```
PikalaxALT: Probably not the best to use the `_MASK` constant here

garakmon: why not?

PikalaxALT: Ironically, it masks what the flags actually represent.

garakmon: so is the solution to replace that constant with the raw value?
and is this a problem everywhere or just here?

PikalaxALT: Create new defines for each collision direction, and a new define that `or`s the 4 directions. Use the new define shifted by `METATILE_COLLISION_MASK`.

garakmon:
okay I updated this but travis had a meltdown for a minute. should be OK if you restart the build tho

PR #668, opened 2018-08-10
macOS compatibility changes

garakmon:
Basically copying my PR from pokeemerald.

PR #756, opened 2019-08-19
Add enum for AI Flags

RiceBoxer:
Cleans up definitions of trainer's AI Flags

Used these files as reference for naming:
data/battle_ai_scripts.s
src/battle_ai_script_commands.c

### DIFF
--- a/include/battle_ai_script_commands.h
+++ b/include/battle_ai_script_commands.h
@@ -16,6 +16,24 @@ enum
     BATTLEAI_DO_NOT_PROCESS
 };
 
+
+
+#define AI_FLAG_CHECK_BAD_MOVE        (1 <<  0)
+#define AI_FLAG_CHECK_VIABILITY       (1 <<  1)
+#define AI_FLAG_TRY_TO_FAINT          (1 <<  2)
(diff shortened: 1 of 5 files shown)


include/battle_ai_script_commands.h
```diff
@@ -16,6 +16,25 @@ enum
     BATTLEAI_DO_NOT_PROCESS
 };
 
+
+enum
```
GriffinRichards: These can probably just be a series of #defines

RiceBoxer: Sure, its changed :)

I wasn't sure which style to go for since the rest of the file used enum and both methods result in identical SHA1s

DizzyEggg:
Why not use Emerald's defines?

PikalaxALT:
@RiceBoxer Please address @DizzyEggg's remark.

PikalaxALT:
This PR will be closed at 21:00 UTC Friday October 11 if the most recent comment is not addressed by then.

coltongit:
I guess since #668 was closed, this could be closed too

PR #757, opened 2019-08-24
English Debug menu translations!

easyaspi314:
I'm back!
```
   　 　          　　　　　　　す
Éß テ-ブ↑& ÒÛªù! (かなテーブルを捨てろよ！/Throw away the kana table!)
```

When compiled in English with DEBUG, instead of óじばÌ (mojibake),
you now get actual English translations!

No more switching from the game, to the kana chart, and if you don't know
Japanese, Google Translate, just to read one sentence.

Additionally, the few unknown translations should use proper encoding or
Romaji.

I decided to use a nicer macro system:

static const u8 sStrHello[] = DTR("こんにちは", "Hello");

This prevents ifdef HELL, as the mixed Japanese and English would
be messy.

Additionally, some special character issues were corrected.

Things untranslated/bugs:
 - Tomomichi's dummy functions. Of course, they don't do much.
 - The title screen battle menu. This uses a Gen 2 style tile rendering
   and the way it is rendered is incredibly overcomplicated.
 - A few phrases which I couldn't get a clear translation for. I don't know
   much Japanese lol.
 - Song names. Just tedious with the limited buffer size.
 - A few layout issues. Some dimensions should be tweaked to avoid text
   box ovetflows.
 - Potential buffer overflows, which I want to try to avoid
   - (ASan for GBA? :morphon:)
 - While I use the DEBUG_TRANSLATE flag (automatic with ENGLISH && DEBUG)
   for C code, the asm data only checks for ENGLISH && DEBUG. Additionally,
   they are ugly.
 - Haven't tested all builds yet.
 - Some code formatting issues.

~~I also made it so this will compile with the system preprocessor (including
Clang), and my agbcc binutils (which will attempt to be selected automatically
if you don't have DKP).

Clang needs a few flags (it can't do asm enums with line numbers and it does
trigraphs in standard mode), but it is no longer flipping out.~~

I am going to restart my build system project. It was getting way off track,
and ucpp was too difficult to maintain (one of the main reasons for the silence).


easyaspi314:
Also, props to TCRF, Jisho, and Google Translate. It got the bulk of the translations.

Makefile
```diff
@@ -1,4 +1,8 @@
+ifneq (,$(DEVKITARM))
```
PikalaxALT: This all conflicts with #742

easyaspi314: Ok I can revert that

PikalaxALT:
@easyaspi314 Please resolve merge conflicts.

PR #763, opened 2019-10-01
Add shasum in addition to sha1sum

GriffinRichards:
Fixes `-bash: sha1sum: command not found` for platforms which use `shasum` instead


Makefile
```diff
@@ -16,7 +16,7 @@ CC1       := tools/agbcc/bin/agbcc$(EXE)
 CPP       := $(PREFIX)cpp
 LD        := $(PREFIX)ld
 OBJCOPY   := $(PREFIX)objcopy
-SHA1SUM   := sha1sum -c
```
PikalaxALT: Or just add `brew install md5sha1sum` to the macOS setup instructions

GriffinRichards: Is there a preference for one change over the other?

GriffinRichards:
Duplicated by #668, close if that's merged

Makefile
coltongit: I think the best possible solution to this would be to let the end-user choose between what's documented here in this PR and `brew install`ing `md5sha1sum`.

GriffinRichards: That will always be the case. This doesn't prevent people from installing `md5sha1sum`

PR #766, opened 2019-10-16
Sync compare behavior with pokeemerald

GriffinRichards:
Turn off compare by default, add rules to use compare. Tossing in ignore the porymap config


Makefile
```diff
@@ -7,6 +7,7 @@ else
 EXE :=
 endif
 
+COMPARE := 0
```
PikalaxALT: Change the line in config.mk instead.

GriffinRichards: Fixed

PR #781, opened 2020-05-17
Sync fieldmap data

GriffinRichards:


GriffinRichards:
I thought about that. Maybe? The assert isn't present here so I used the name pokeemerald uses currently, and I sort of prefer that name anyway

PokeCodec:
Is this pr alive

PR #790, opened 2020-06-08
Unnest if statements

PokeCF:
Overall, it is generally considered bad style to continually nest loops. However, if this is what the decompiler outputted, then why bother touching it?

However, this can be rewritten as returning early if the playtime is less than 60.

This code look looks better, has the same output, and is easier to understand.

### DIFF
--- a/src/play_time.c
+++ b/src/play_time.c
@@ -33,33 +33,33 @@ void PlayTimeCounter_Stop()
     sPlayTimeCounterState = STOPPED;
 }
 
-void PlayTimeCounter_Update()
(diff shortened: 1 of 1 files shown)


PikalaxALT:
Please observe the following etiquette when proposing changes:
- Set up the project on your local machine according to [INSTALL.md](INSTALL.md).
- Bundle proposed changes into a single PR, rather than opening multiple small PRs.
- Do not use the leaked official source code as a reference. Doing so disqualifies you from contributing to this effort.

PR #792, opened 2020-06-09
Replace sub813CE88 with function name

PokeCF:
sub_813CE88 is responsible for fetching the sprites of pokemon and setting them up for the into animation, returning a spriteID.

So I renamed the function to better reflect that.

If you have any better ideas, please feel free to critique.


PikalaxALT:
Take the hint.

PR #793, opened 2020-07-07
Match SortPokedex

PokeCodec:
The assembly for comparing r5 is lsr	r5, r1, #0x18, so having that be a bool32 makes no sense. After I set r5 to the same type as i, and moved the variables to the top to prevent the cross jump optimization, I found the asm was no longer needed!


GriffinRichards:
This doesn't match

PokeCodec:
Fixed it!

PR #794, opened 2020-07-14
Match sub_80F1BC8

PokeCodec:
The original implementation of this function works, except for one major oversight: one asm instruction in the code, " beq_080F1DC4" branch if equal. Now, this looked strange to me, as there was only a branch if r3 was NOT equal in the nonmatching code. I looked for the label in the asm and found it was going to an area in which the function was returning false. I did that, and the code matched up better, except for some registers, so I moved the variables on the top within a scope and found that it matched!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: generic
-dist: trusty
+dist: bionic
 sudo: false
 env:
   global:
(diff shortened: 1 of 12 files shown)


GriffinRichards:
This was merged incorrectly and has unrelated changes

PikalaxALT:
Lack of activity indicates an unwillingness to fix the indicated changes. Author is advised to open this PR anew with the above indicated issues resolved.

PR #796, opened 2020-08-08
Match More Functions

PokeCodec:


PikalaxALT:
PR title does not describe work performed

PokeCodec:
@PikalaxALT fixed

src/pokedex.c
```diff
@@ -5084,74 +5084,24 @@ static void sub_80927F0(u8 taskId)
...
 
     for (i = 0; i < d; i++)
     {
-        ((u16 *)VRAM)[15 * 0x400 + c * 32 + i + b] &= 0xFFF;
-        ((u16 *)VRAM)[15 * 0x400 + c * 32 + i + b] |= a << 12;
+        j = *(u16 *)(VRAM + 15 * 0x800 + c * 64 + (b + i)*2);
+        j &= 0xFFF;
+        j |= a << 12;
+
+        *(u16 *)(VRAM + 15 * 0x800 + c * 64 + (b + i)*2) = j;
```
PikalaxALT, line 5097: This looks macro-like

PokeCodec: Should I make a macro?

PokeCodec: I don't think a macro is needed because macros are for reuse. this is the only time I do this.

src/pokedex.c
```diff
@@ -25,7 +25,7 @@
 #include "trig.h"
 #include "scanline_effect.h"
 #include "ewram.h"
-
+#define VRAM_ADDR(a, b, c) *(u16 *)(BG_VRAM + (15 * 0x800) + (c) * 64 + ((b) + (a))*2)
```
GriffinRichards, line 28: Make this define once, don't duplicate it in every file it's needed

PokeCodec:
Each and every issue addressed!

constants/m4a_constants.inc
```diff
@@ -1,199 +1,183 @@
-	.equiv ID_NUMBER, 0x68736d53
+.equiv ID_NUMBER, 0x68736d53
```
PikalaxALT, line 1: Revert changes in m4a_constants.inc

include/gba/m4a_internal.h
```diff
@@ -140,55 +140,13 @@ struct SoundChannel
     struct WaveData *wav;
     u32 cp;
     struct MusicPlayerTrack *track;
-    u32 pp;
-    u32 np;
-    u32 d4;
+    void* pp;
```
PikalaxALT, line 143: Pointer style inconsistent with line 103

include/gba/m4a_internal.h
```diff
@@ -334,7 +292,47 @@ struct Song
 extern const struct MusicPlayer gMPlayTable[];
 extern const struct Song gSongTable[];
 
+#define MAX_DIRECTSOUND_CHANNELS 12
```
PikalaxALT, line 295: Why move this definition?

src/battle/anim/psychic.c
```diff
@@ -516,7 +516,7 @@ static void sub_80DB92C(struct Sprite *sprite)
 {
     REG_BLDALPHA = ((16 - sprite->data[3]) << 8) | sprite->data[3];
 
-    if (--sprite->data[3] == -1)
+    if (sprite->data[3]-- == 0)
```
PikalaxALT, line 519: Is this change required to match?

src/libs/libagbsyscall.s
```diff
@@ -5,61 +5,61 @@
 
 	thumb_func_start ArcTan2
 ArcTan2: @ 81E07E0
-	swi 0xA
+	svc 0xA
```
PikalaxALT, line 8: Revert

src/palette.c
```diff
@@ -149,47 +149,43 @@ void ReadPlttIntoBuffers(void)
 bool8 BeginNormalPaletteFade(u32 selectedPalettes, s8 delay, u8 startY, u8 targetY, u16 blendColor)
 {
     u8 temp;
-    register u32 _blendColor asm("r8") = blendColor;
 
-    if (gPaletteFade.active)
+    if (gPaletteFade.active == TRUE)
```
PikalaxALT, line 153: ?????

PokeCodec: if (a == true) is NOT the same as if (a). they generate different code.

include/gba/m4a_internal.h
PokeCodec: So it can be grouped together with the PCM_DMA_BUF_SIZE, as both macros are used in conjunction a lot

src/battle/anim/psychic.c
PokeCodec: It makes more sense this way

PokeCodec:
Fixed!

PokeCodec:
if (a == true) is NOT the same as if (a). they generate different code. @PikalaxALT

PR #788, opened 2020-05-26
Fix NONMATCHINGs and actual English Debug support

kazblox:
A big one. This will take long to review, but I am trying to get pokeruby up into at least a better state and parity with pokeemerald after seeing the sources unfortunately leak this early.

Most of it is multiple MATCHINGS ported from pokeemerald and pokefirered. Fixes unique to this PR however include:
- sub_80D4D64 _(new behavior? parameter poking and local temp combination)_
- **AnimTask_CreateSurfWave** _(GF used DMA3FillLarge32)_
- **_atk24_** _(permutation and bitfield representation issue)_
- CopyWallpaperTilemap, formerly sub_8099F58
- sub_8106B90 _(matched with raw pointer arithmetic instead of arrays, also parenthesis/grouping memes)_
- sub_80D3D68 _(grouping memes)_
Of which I will port over to pokeemerald/firered when merged, if compatible.

English Debug support using the ROMs from the leak is also underway.

PR #797, opened 2020-08-08
Match sub_802DB6C

PokeCodec:


GriffinRichards:
This duplicates work in #788

PokeCodec:
How was I supposed to know that? That's a HUGE PR.

GriffinRichards:
I'm telling you now that it does

PR #807, opened 2020-10-24
[EMERALD_INFORMED] Match InitObjectEventStateFromTemplate

MeatLoaf3:
This match was done via copy-pasting pokeemerald.


PikalaxALT:
Requester is blacklisted

PR #808, opened 2020-10-25
Paste dewford trend from emerald

MeatLoaf3:
It's matched in pokeemerald, so I copied from the pret emerald repo and pasted it here


MeatLoaf3:
I forgot to change gSaveBlock1Ptr to gSaveBlock1 but now I changed it back to gSaveBlock1 and got OK.

PikalaxALT:
Requester is blacklisted

PR #809, opened 2020-10-25
Copy FadeOutBody from Emerald

MeatLoaf3:
Copied from pokeemerald. Original match by Pidgey and Kurausukun. Leads to OK.


PikalaxALT:
Requester is blacklisted

PR #819, opened 2021-01-15
Fix fakematch in battle_transition.c

PokeCodec:
I don't know why the var isn't s16 to start, but GF will GF.

Taken from pokeemerald's matching of the same function.


PokeCodec:
Any updates?

PR #822, opened 2021-04-28
Name map groups

GriffinRichards:


data/maps/map_groups.json
```diff
@@ -1,41 +1,41 @@
...
-    "gMapGroup28",
-    "gMapGroup29",
-    "gMapGroup30",
-    "gMapGroup31",
-    "gMapGroup32",
-    "gMapGroup33"
-  ],
-  "gMapGroup0": [
+    "gMapGroup_TownsAndRoutes",
+    "gMapGroup_IndoorLittleroot",
```
huderlem, line 4: `Indoors` instead of `Indoor` sounds better in my head

GriffinRichards: Seems we went with `Indoor` for the other repos

PR #842, opened 2022-04-28
use json for region map sections

garakmon:


coltongit:
This one probably should be a draft because of it's companion in pokeemerald not finished yet (https://github.com/pret/pokeemerald/pull/1651)

PR #848, opened 2022-07-15
Initial Features

phi1997:
Have player and rivals use Spheal as starter, with dialogue and opponent rosters reflecting this. Add minor start of game QoL changes.


phi1997:
Oops, meant to be only on my fork

PR #875, opened 2025-02-26
All 16 towns and cities updated for data name parity with pokeemerald.

xyifer12:
CHANGES: Names for EventScripts, FLAGs, Text have been modified to match those of PokeEmerald as closely as possible where applicable. New names have been created where needed. 

Petalburg, Slateport, Mauville, Rustboro, Fortree, Lilycove, Mossdeep, Sootopolis, EverGrande, Littleroot, Oldale, Dewford, Lavaridge, Fallarbor, Verdanturf, Pacifidlog all modified. 

Other text and script files have been modified where necessary for changes to be made in town and route files, this includes berries.inc and event_scripts.s for example.
- - -
TESTING DONE: "make [version] COMPARE=1" for ruby, ruby_rev1, ruby_rev2, ruby_debug, sapphire, sapphire_rev1, sapphire_rev2, sapphire_debug.
"make [version] COMPARE=1" for ruby_de, ruby_de_debug, sapphire_de, sapphire_de_debug as well.
- - -
NOTES: I've touched *most* but not all names. For some, I couldn't find or decide on accurate names. For others, I just didn't want to deal with the specific data. I've checked this in every version of the game multiple times, but it could *really* use another set of eyes looking for oversights and mistakes.


data-de/event_scripts.s
```diff
@@ -859,7 +859,7 @@ Common_EventScript_NoRoomForDecor:: @ 81A02B8
 	msgbox Text_NoRoomLeftForAnother, MSGBOX_DEFAULT
 	return
 
-Common_EventScript_SetLegendaryWeather:: @ 81A02C1
+Common_EventScript_SetAbnormalWeather:: @ 81A02C1
```
GriffinRichards, line 862: The old name is fine here, the `Abnormal` from pokeemerald refers to `WEATHER_ABNORMAL`, which is an Emerald-exclusive weather that alternates between drought and heavy rain.

data/maps/FallarborTown/scripts.inc
```diff
@@ -1,55 +1,55 @@
...
 	end
 
-FallarborTown_EventScript_14E842:: @ 814E842
-MeteorFalls_1F_1R_EventScript_14E842:: @ 814E842
-MeteorFalls_1F_2R_EventScript_14E842:: @ 814E842
-	msgbox FallarborTown_Text_16CA83, MSGBOX_SIGN
+FallarborTown_EventScript_TownSign:: @ 814E842
+MeteorFalls_1F_1R_EventScript_TownSign:: @ 814E842
+MeteorFalls_1F_2R_EventScript_TownSign:: @ 814E842
+	msgbox FallarborTown_Text_TownSign, MSGBOX_SIGN
```
GriffinRichards, line 50: It doesn't need to change right now, but when you hit cases like this in general they should be collapsed into a single label, i.e.:
```diff
FallarborTown_EventScript_TownSign:: @ 814E842
-MeteorFalls_1F_1R_EventScript_TownSign:: @ 814E842
-MeteorFalls_1F_2R_EventScript_TownSign:: @ 814E842
	msgbox FallarborTown_Text_TownSign, MSGBOX_SIGN
```
and uses of `MeteorFalls_1F_1R_EventScript_TownSign` or `MeteorFalls_1F_2R_EventScript_TownSign` would be replaced with `FallarborTown_EventScript_TownSign`

data/maps/LittlerootTown/scripts.inc
```diff
@@ -1,125 +1,125 @@
...
-LittlerootTown_EventScript_14D563:: @ 814D563
+LittlerootTown_EventScript_SetRivalLeftForRoute103:: @ 814D563
 	setflag FLAG_RIVAL_LEFT_FOR_ROUTE103
 	return
 
-LittlerootTown_EventScript_14D567:: @ 814D567
+LittlerootTown_EventScript_HideMapNamePopup:: @ 814D567
 	setvar VAR_OLDALE_STATE, 2
 	setflag FLAG_HIDE_RIVAL_OLDALE_TOWN
 	return
```
GriffinRichards, line 30: This is pokeemerald's `LittlerootTown_EventScript_MoveRivalFromOldale`, not `LittlerootTown_EventScript_HideMapNamePopup`

xyifer12: Good catch, thank you. I'm working on implementing and testing revisions now.

data/maps/FallarborTown/scripts.inc
xyifer12: I was under the impression that this would interfere with matching, good to know!

